PR #1, opened 2020-07-29
ChipWhisperer Initial Capture Scripts

moidx:
Capture scripts demonstrating how to generate captures with OpenTitan running on a CW305 Artix-7 100T FPGA board. The captures are generated using the ChipWhisperer Simple Serial test interface via UART. See doc/getting_started.md for more details.


cw/cw305/capture_traces.py
```diff
@@ -0,0 +1,87 @@
...
+
+def initialize_test(device_cfg):
+  """Initialize test setup."""
+  ot = device.OpenTitan(
+    device_cfg['fpga_bitstream'],
+    device_cfg['fw_bin'],
+    device_cfg['pll_frequency'],
+    device_cfg['baudrate'])
+  ot.initialize()
+  print('Scope setup with sampling rate {} S/s'.format(ot.scope.clock.adc_rate))
```
alphan, line 27: How about using `f''` instead of `''.format()`, i.e. `f'Scope setup with sampling rate {ot.scope.clock.adc_rate} S/s'`?

cw/cw305/capture_traces.py
```diff
@@ -0,0 +1,87 @@
...
+  return ot
+
+
+# Based on https://github.com/newaetech/chipwhisperer-jupyter/blob/master/PA_HW_CW305_1-Attacking_AES_on_an_FPGA.ipynb
```
alphan, line 31: Can move into the docstring.

cw/cw305/util/device.py
```diff
@@ -0,0 +1,84 @@
...
+      self.bitstream = bitstream
+      self.fw_image = fw_image
+      self.pll_frequency = pll_frequency
+      self.initialized = False
+      self.fpga = None
+      self.scope = None
+      self.target = None
+      self.baudrate = baudrate
+
+  def initialize(self):
```
alphan, line 27: Is there a reason why this can not be inlined in `__init__`?

cw/cw305/util/device.py
```diff
@@ -0,0 +1,84 @@
...
+class OpenTitan(object):
+  def __init__(self, bitstream, fw_image, pll_frequency, baudrate):
+      self.bitstream = bitstream
+      self.fw_image = fw_image
+      self.pll_frequency = pll_frequency
+      self.initialized = False
+      self.fpga = None
+      self.scope = None
+      self.target = None
+      self.baudrate = baudrate
```
alphan, line 25: Consider adding these as arguments to `initialize_*` methods below and calling them directly here. They can return the objects that they initialize instead of working on the attributes of `self`.

cw/cw305/capture.yaml
```diff
@@ -0,0 +1,12 @@
...
+  fpga_bitstream: objs/lowrisc_systems_top_earlgrey_cw305_0.1.bit
+  fw_bin: objs/aes_serial_fpga_nexysvideo.bin
+  pll_frequency: 25000000
+  baudrate: 57600
+capture:
+  # Only AES-128 ECB is supported at this moment.
+  key_len_bytes: 16
+  plain_text_len_bytes: 16
+  num_traces: 5000
+  project_file: projects/opentitan_simple_aes.cwp
```
alphan, line 11: Do we need the suffix? (according to https://chipwhisperer.readthedocs.io/en/latest/api.html, we don't) If not, consider renaming this to `project` or `project_name`.

cw/cw305/capture_traces.py
moidx: Done. Updated all strings to use the suggested format approach.

cw/cw305/util/device.py
moidx: Updated to follow the approach you suggested below.

cw/cw305/util/device.py
moidx: Done.

cw/cw305/capture.yaml
moidx: Updated `project_name`

cw/cw305/capture_traces.py
```diff
@@ -0,0 +1,87 @@
...
+  # Key and plaintext generator
+  ktp = cw.ktp.Basic()
+  ktp.key_len = capture_cfg['key_len_bytes']
+  ktp.text_len = capture_cfg['plain_text_len_bytes']
+  ot.target.output_len = capture_cfg['plain_text_len_bytes']
+
+  key, text = ktp.next()
+
+  cipher = AES.new(bytes(key), AES.MODE_ECB)
+  print(f'Using key: {binascii.b2a_hex(bytes(key))}')
```
alphan, line 45: Minor: If a hex string is acceptable, consider using `hex()` method of a `bytes` object, i.e. `bytes(key).hex()`. Applies to other usages as well.

moidx:
@vogelpi  I couldn't rebase and merge using the GUI due to GitHub not being able to sign commits in the web interface. I ended up having to install the https://hub.github.com/ CLI tooI and do the rebase and merge operations from the command line. 

I am not sure if there is a better way to do this.

alphan:
> @vogelpi I couldn't rebase and merge using the GUI due to GitHub not being able to sign commits in the web interface. I ended up having to install the https://hub.github.com/ CLI tooI and do the rebase and merge operations from the command line.
> 
> I am not sure if there is a better way to do this.

You can also rebase locally, push that commit after signing and merge from here. Not sure which one is simpler.

vogelpi:
Sorry, I did not manage to review this in time.

Regarding your point on the rebasing: Normally I also do the rebase locally and then force push to the PR as suggested by @alphan . I have never used the hub CLI tool. Is there any benefit in using it for this task? Sorry, I might not fully understand something here.

moidx:
My branch was rebased, but somehow the rebase and merge button failed.

I think the issue is that rebase within GitHub always triggers commit hash updates, which can't be signed in the web interface. https://docs.github.com/en/github/administering-a-repository/about-merge-methods-on-github#rebasing-and-merging-your-commits

I ended up having to get the pull request commits into a local instance of lowrisc/ot-sca to do a manual push. I think this is probably the reason lowrisc/opentitan doesn't have signed commits enabled and only rely on a CI commit lint check?

vogelpi:
Ah, now I see the problem. I am very sorry @moidx . I did not mean to enforce commit signing with verified signatures. Just wanted to have the sign-off line as we have it in OpenTitan. But I now remember that we just use CI to check this over there. I have now disabled the check for ot-sca.

PR #2, opened 2020-08-04
Add configuration options for spiflash

moidx:
* Make spiflash configuratble from capture.yaml.
* Add FTDI connectivity instructions to the doc/getting_started.md
  guide.

### DIFF
--- a/cw/cw305/capture_traces.py
+++ b/cw/cw305/capture_traces.py
@@ -14,13 +14,20 @@ import chipwhisperer as cw
 
 from util import device
 from util import plot
+from util import spiflash
 
 
-def initialize_test(device_cfg):
-  """Initialize test setup."""
+def initialize_capture(device_cfg, spiflash_cfg):
+  """Initialize capture."""
+  fw_programmer = spiflash.FtdiProgrammer(
+    spiflash_cfg['bin'],
+    spiflash_cfg['dev_id'],
+    spiflash_cfg['dev_sn'],
+    device_cfg['fw_bin'])
+
   ot = device.OpenTitan(
+    fw_programmer,
     device_cfg['fpga_bitstream'],
-    device_cfg['fw_bin'],
     device_cfg['pll_frequency'],
     device_cfg['baudrate'])
   print(f'Scope setup with sampling rate {ot.scope.clock.adc_rate} S/s')
@@ -83,5 +90,5 @@ def run_capture(capture_cfg, ot):
 if __name__ == "__main__":
   with open('capture.yaml') as f:
     cfg_file = yaml.load(f, Loader=yaml.FullLoader)
-  ot = initialize_test(cfg_file['device'])
+  ot = initialize_capture(cfg_file['device'], cfg_file['spiflash'])
   run_capture(cfg_file['capture'], ot)
--- a/cw/cw305/util/spiflash.py
+++ b/cw/cw305/util/spiflash.py
@@ -1,0 +1,28 @@
+# Copyright lowRISC contributors.
+# Licensed under the Apache License, Version 2.0, see LICENSE for details.
+# SPDX-License-Identifier: Apache-2.0
+"""Utilities used to download firmware images into the OpenTitan system."""
+
+import subprocess
+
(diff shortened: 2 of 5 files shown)


cw/cw305/util/spiflash.py
```diff
@@ -0,0 +1,27 @@
+# Copyright lowRISC contributors.
+# Licensed under the Apache License, Version 2.0, see LICENSE for details.
+# SPDX-License-Identifier: Apache-2.0
+r"""Utilities used to download firmware images into the OpenTitan system."""
```
alphan, line 4: I don't think this should be a raw string.

cw/cw305/util/spiflash.py
```diff
@@ -0,0 +1,27 @@
...
+  See doc/getting_started.md for details on how to connect the FTDI cable to
+  the CW305 FPGA target board.
+  """
+  def __init__(self, spiflash_path, dev_id, dev_sn, input):
+    """Initializes spiflash FTDI programmer.
+
+    @param spiflash_path Path to spiflash executable.
+    @param dev_id FTDI USB device ID in vendor:product hex format.
+    @param dev_sn FTDI USB serial number as reported by lsusb.
+    @param input Path to OpenTitan firmware image.
```
alphan, line 20: Are we using doxygen for python as well? If not, we should probably follow [this](https://google.github.io/styleguide/pyguide.html#383-functions-and-methods).

cw/cw305/capture_traces.py
```diff
@@ -14,13 +14,20 @@
 
 from util import device
 from util import plot
+from util import spiflash
 
 
-def initialize_test(device_cfg):
+def initialize_test(device_cfg, spiflash_cfg):
```
alphan, line 20: Maybe this should be called `initialize_capture` for the sake of clarity (+ the docstring below).

cw/cw305/util/spiflash.py
```diff
@@ -0,0 +1,27 @@
...
+    """
+    self.cmd = [spiflash_path, f'--dev-id={dev_id}', f'--dev-sn={dev_sn}',
+                f'--input={input}']
+
+  def run(self):
+    """Run spiflash utility with predifined command line arguments."""
```
alphan, line 26: typo: predifined

doc/getting_started.md
```diff
@@ -7,6 +7,26 @@ In the meantime, check the following PRs:
 * https://github.com/lowRISC/opentitan/pull/2587
 * https://github.com/lowRISC/opentitan/pull/2735
 
+### Downloading code via FTDI
+
+Currently the target firmware is downloaded using the
+[spiflash](https://docs.opentitan.org/sw/host/spiflash/README/) utility. A
+debian-compatible pre-built binary is available at cw/cw305/bin/linux/spiflash.
```
vogelpi, line 14: I understand that it is convenient to "ship" this and other binaries (bitstreams) directly with the code in the same repo. I also think this is okay now that we are at the very beginning as it eases collaboration and allows to get up to speed more easily.

However, I don't think we should keep updating these binaries. Some of them are pretty big and they will remain in the tree forever. Once we have things up and running, I suggest we kick the binaries out again and create a script to download them from GitHub releases. There we can periodically provide new binaries as we do with the toolchains https://github.com/lowRISC/lowrisc-toolchains/releases.

moidx: SGTM. That's a good idea :)

cw/cw305/util/spiflash.py
moidx: Done.

cw/cw305/util/spiflash.py
moidx: SGTM. I prefer the format you suggested.

I will let lowRISC weigh on this for additional coding style guidelines across projects.

cw/cw305/capture_traces.py
moidx: Done.

cw/cw305/util/spiflash.py
moidx: Fixed.

PR #4, opened 2020-08-14
Use Git Large File Storage (LFS) for all binaries

vogelpi:
This PR changes the repo to use Git Large File Storage (LFS) for storing binary files. Previously, these binaries were stored in tree. This should be avoided as binary diffs are not possible and Git will always store the entire binary file for all times. For this reasons, the size of the repo will grow very quickly and whenever doing a `git clone` the entire history including all versions of all binaries is downloaded.

With Git LFS, these binaries are stored on a different remote server and the repo just contains diff-able text pointers to the binaries. By using the `git-lfs` command line tool, the whole process is completely transparent to the user.


vogelpi:
Thanks for your feedback @alphan . Good catch about `doc/sample_traces.png`, I will also move that over. Even though we won't update it frequently.

I wasn't aware that storage cannot be reclaimed with Git LFS and I agree that once we share other stuff like traces we should rethink this.

moidx:
WOW, it is unfortunate that git-lfs  has this limitation.  I agree, we can setup a Google storage bucket or something similar to share traces down the line.

colinoflynn:
Yeah power traces in GIT even with git-lfs will be very un-fun, so
recommend other things ;-)

FWIW - we're working on using zarr + some GCP buckets for some (unrelated)
power trace sharing.

alphan:
Thanks for mentioning `zarr` @colinoflynn, it looks really interesting. A very superficial comparison between `dvc` and `zarr`:
* `zarr` basically abstracts away the underlying storage and supports a wide variety of formats/services: zip, db, S3, GCP, Azure, etc. It looks like it doesn't support versioning but [this may change](https://github.com/zarr-developers/zarr-specs/issues/76).
* `dvc` is more like a data pipeline building tool that also supports versioning large datasets (using `git`) stored in some cloud service: can define pipelines, run experiments, and compare results.

It looks like the main difference between the two is, if using `dvc` you don't even have to mention its name in your code (even though you can if you choose to, it has a python API) and you will have a more ergonomic workflow (when compared to `make`), but if using `zarr` you have to use its APIs in your code.

Interesting stuff, I'll keep [reading](https://towardsdatascience.com/why-git-and-git-lfs-is-not-enough-to-solve-the-machine-learning-reproducibility-crisis-f733b49e96e8).

PR #5, opened 2020-08-15
[objs] Update bitstream and AES tests

moidx:
Update bistream and AES test with new trigger. The new trigger is
controlled by the AES idle signal for better trace alignment.

Updated capture configuration to take into account new triggger setup.


cw/cw305/util/device.py
```diff
@@ -23,7 +23,7 @@ def initialize_fpga(self, bitstream, pll_frequency):
     """Initializes FPGA bitstream and sets PLL frequency."""
     print('Connecting and loading FPGA')
     fpga = cw.capture.targets.CW305()
-    fpga.con(bsfile=bitstream, force=True)
+    fpga.con(bsfile=bitstream, force=False)
```
alphan, line 26:
```suggestion
    # Do not program the FPGA if it is already programmed.
    fpga.con(bsfile=bitstream, force=False)
```

moidx: Done.

PR #7, opened 2020-10-07
Update config and binaries for latest changes in OpenTitan

vogelpi:
The main clock of the target is now 10 MHz and we use that directly for the ADC on the capture board.


vogelpi:
The binaries included in this PR are much more current but some of the changes are not yet in OpenTitan master.

Regarding the `spiflash` binary: I can't run this as I have the wrong glibc installed here (Ubuntu GLIBC 2.27-3ubuntu1.2 vs. GLIBCXX_3.4.26). Can we include a statically linked version of `spiflash`?

cw/cw305/capture.yaml
```diff
@@ -1,8 +1,8 @@
 device:
   fpga_bitstream: objs/lowrisc_systems_top_earlgrey_cw305_0.1.bit
   fw_bin: objs/aes_serial_fpga_nexysvideo.bin
-  pll_frequency: 25000000
-  baudrate: 57600
+  pll_frequency: 100000000
```
moidx, line 4: nice!

moidx (approved):
LGTM. 

I will look at statically compiling spiflash.

alphan (approved):

vogelpi:
> I will look at statically compiling spiflash.

That would be awesome!

Thanks guys for the approval.

moidx:
Lets check this in as is. 

I think we need to discuss in more details the implications of maintaining a static library, and alternative options.

PR #8, opened 2020-10-13
Mix columns attack

moidx:
Add mix_column HD attack. The attack leverages the mix column attack model implemented in the ChipWhisperer library. See Python module documentation for more details.

### DIFF
--- a/cw/cw305/mix_columns_capture.py
+++ b/cw/cw305/mix_columns_capture.py
@@ -1,0 +1,40 @@
+#!/usr/bin/env python3
+# Copyright lowRISC contributors.
+# Licensed under the Apache License, Version 2.0, see LICENSE for details.
+# SPDX-License-Identifier: Apache-2.0
+"""Mix column HD Attack - Capture portion
+
+Attack implemented by ChipWhisperer:
+Repo: https://github.com/newaetech/chipwhisperer-jupyter/blob/master/experiments/MixColumn%20Attack.ipynb
+Reference: https://eprint.iacr.org/2019/343.pdf
+
+See mix_columns_cpa_attack.py for attack portion.
+"""
+import chipwhisperer as cw
+import yaml
+
+import capture_traces as cp
+
+if __name__ == "__main__":
+  with open('capture.yaml') as f:
+    cfg_file = yaml.load(f, Loader=yaml.FullLoader)
+  ot = cp.initialize_capture(cfg_file['device'], cfg_file['spiflash'])
+  ot.target.output_len = cfg_file['capture']['plain_text_len_bytes']
+
+  # Key and plaintext generator
+  ktp = cw.ktp.VarVec()
+  ktp.key_len = cfg_file['capture']['key_len_bytes']
+  ktp.text_len = cfg_file['capture']['plain_text_len_bytes']
+
+  project_name = cfg_file['capture']['project_name']
+
+  # For each iteration, run a capture where only the bytes specified in
+  # `text_range` are set to random values. All other bytes are set to a
+  # fixed value.
+  for var_vec in range(4):
+    cfg_file['capture']['project_name'] = f'{project_name}_{var_vec}'
+    ktp.var_vec = var_vec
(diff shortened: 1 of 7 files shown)


moidx:
The attack is still not perfect. Currently investigating.

Results for 5K traces per iteration (4 iterations total):

```console
known_key: b'2b7e151628aed2a6abf7158809cf4f3c'                                                                                                                          
key guess: b'1a7e1516ccaed2a668f7158812cf4f3c'
```

cw/cw305/mix_columns_capture.py
```diff
@@ -0,0 +1,40 @@
...
+import capture_traces as cp
+
+if __name__ == "__main__":
+  with open('capture.yaml') as f:
+    cfg_file = yaml.load(f, Loader=yaml.FullLoader)
+  ot = cp.initialize_capture(cfg_file['device'], cfg_file['spiflash'])
+  ot.target.output_len = cfg_file['capture']['plain_text_len_bytes']
+
+  # Key and plaintext generator
+  ktp = cw.ktp.VarVec()
```
vogelpi, line 25: This should be `cw.ktp.Basic()` I think. At least there is no `VarVec()` in ChipWhisperer v.5.3.1.

cw/cw305/mix_columns_cpa_attack.py
```diff
@@ -0,0 +1,33 @@
+#!/usr/bin/env python3
+# Copyright lowRISC contributors.
+# Licensed under the Apache License, Version 2.0, see LICENSE for details.
+# SPDX-License-Identifier: Apache-2.0
+import binascii
+import chipwhisperer as cw
+from chipwhisperer.analyzer.attacks.attack_mix_columns import AttackMixColumns
```
vogelpi, line 7: In ChipWhisperer v.5.3.1 there is no `attack_mix_columns` in `chipwhisperer.analyzer.attacks`. Can you please confirm the version number @moidx? Maybe we need to update the `python-requirements.txt`.

cw/cw305/util/device.py
```diff
@@ -42,9 +42,9 @@ def initialize_fpga(self, bitstream, pll_frequency):
   def initialize_scope(self):
     """Initializes chipwhisperer scope."""
     scope = cw.scope()
-    scope.gain.db = 25
+    scope.gain.db = 30
     # Samples per trace - We oversample by 10x and AES is doing ~12/16 cycles per encryption.
-    scope.adc.samples = 180
+    scope.adc.samples = 200
```
vogelpi, line 47: Do we really need more samples? I would assume that there is no useful information in the last 20 samples. Everything useful for the attack should happen during the first 120 (AES-128) - 160 (AES-256) samples max.

vogelpi (commented):
Thanks @moidx for working on this! The code looks good but I can't run it on my end. There are some incompatibilities with the current version 5.3.1 of the ChipWhisperer API. Can you please confirm the exact version you are using? We can then fix that in the python requirements.

cw/cw305/mix_columns_cpa_attack.py
moidx: Hi @vogelpi, I used the git installation option on chipwhisperer to be able to get the latest changes from their `develop` branch: https://chipwhisperer.readthedocs.io/en/latest/installing.html#install-repo-git

I created a fork of the repo and used an upstream configuration similar to what we do with OpenTitan repos.

cw/cw305/mix_columns_capture.py
moidx: See comment below.

cw/cw305/util/device.py
moidx: Probably not. I will revert this change.

vogelpi: Thanks

cw/cw305/mix_columns_cpa_attack.py
vogelpi: Ah I see! Thanks for providing that info. Now the attack is also working in my setup.

Issue #14, opened 2020-10-19
Restructure the repo directories


moidx:
Currently, data capture, target and analysis scripts are stored in the same directory.  Consider a top level view similar to the following:

```console
tree .
.
├── analysis (models, pre/post processing libraries)
├── attacks (mountable attacks)
├── capture (data capture)
│   ├── cw
│   └── picoscope
├── obj (access to precompiled bitstreams, etc, TBD)
├── targets (target boards)
│   └── cw305
├── tests (integration tests, and end-to-end testing)
└── utils (utility libraries, e.g. cw to trs conversion scripts)
```

johannheyszl:
as discussed in [doc](https://docs.google.com/document/d/1HjsIG-4aXvQZqbH-zv8tTlzKCQEE1RdwtXapCEydKFU/edit#bookmark=id.8kucy0hjm4ea)

PR #17, opened 2020-10-20
New dir structure

moidx:
Implements directory structure proposed in lowrisc/ot-sca#14. The Python root folder is named `otsca` to make clear the use of local dependencies in the package.


moidx:
@vogelpi, @alphan  I added this to get feedback on the directory structure. PLMK your thoughts. Thanks!

vogelpi (commented):
Thanks @moidx for the PR. It's a good idea to restructure the repo. But the current structure is not fully clear to me and it is also not fully aligned with your proposal in #14 .

For example here we have:
```
otsca
├── attacks
│   ├── mix_columns
│   │   ├── capture.py
│   │   └── cpa_attack.py
│   └── simple
│       ├── capture.py
│       └── cpa_attack.py
├── capture
│   └── cw
│       └── open_adc.py
├── config
│   └── capture.yaml
├── targets
│   └── cw305
│       └── device.py
└── util
    ├── plot.py
    └── spiflash.py

```
Meaning we have stuff labeled with "capture" in three different locations. IMO, the `attack` folder should really only contain the attacks (the `mix_columns` is a bit special as it requires special capture) but I would assume most attacks share the same simple capture script. Anyway this is not so critical if we can reduce the usage of the term capture a bit (see next paragraph).

Maybe the `capture` folder should be named `scope` or `scopes` (to be consistent with `targets`) as it currently just contains the script to configure the ADC on the ChipWhisperer Lite and according to #14 we might eventually end up with a second subfolder `picoscope` in there. Maybe `cw` should be renamed to `cw-lite` to be consistent (under targets we are also using `cw305` and not just `cw`).

What is also a bit confusing is that the only config file we have is `capture.yaml`. This should probably be called `cw` because it contains the ChipWhisperer config for the capture scripts. There might follow other configs maybe. WDYT?

Also, the name of the `otsca` folder is a bit inconsistent (the repo is called `ot-sca`) and the reason for putting all python code in there is not obvious. Maybe we should simply rename it to `py`?

alphan:
Thanks for thinking about the directory structure and creating this PR @moidx! I think this is definitely a step in the right direction. Obviously there are many ways to approach this. Here are some things that we can also consider going forward:
* It would be nice to have a clear separation between libraries and scripts that are run directly. The former can live in a folder named like `lib` and the latter can live in a folder like `attacks` out of the lib folder. If we find that an attack needs to be reused as part of another attack, then it can become a part of the lib obviously (similar to what you have here) but we would still have another script that is run directly.
* This is somewhat minor but instead of modifying `PYTHONPATH`, we can consider modifying `sys.path` in scripts that are run directly.
* I think config file should be out of the lib folder, since it depends on the target and capture setup.
* If we plan to add FI related functionality, we should probably come up with a more general name.

> Meaning we have stuff labeled with "capture" in three different locations. IMO, the attack folder should really only contain the attacks (the mix_columns is a bit special as it requires special capture) but I would assume most attacks share the same simple capture script. Anyway this is not so critical if we can reduce the usage of the term capture a bit (see next paragraph).

Each attack involves programming the target, acquisition, processing, (some analysis during implementation), and the actual attack. IMHO splitting this process into individual scripts is clearer and having separate capture scripts can actually be useful, e.g. when you want to capture once but run the attack multiple times during development or just want to see what the signals look like. These scripts can also handle metadata as well as storage and retrieval.

@vogelpi, would it be OK to split your feedback into two sets, i.e. things that should be addressed in this PR and things that could potentially take more time or could be addressed later? We can move the latter to the issue to address them incrementally.

vogelpi:
Thanks @alphan for the review and clarification around the capture script. Good suggestion to split out library code into a separate lib folder, I like it!

I am fine with splitting up my feedback into two sets. I don't know if it really makes things easier to split it. IMO doing this now as we have less code would be the most efficient way. But I definitely don't want to block things here.

alphan:
> I am fine with splitting up my feedback into two sets. I don't know if it really makes things easier to split it. IMO doing this now as we have less code would be the most efficient way. But I definitely don't want to block things here.

@vogelpi, sure! Apologies for not being clear. I definitely agree w/ the renames that you suggested and IMHO they are quick fixes (up to @moidx of course). I just thought your comments about the use of `capture` in the names and the `otsca` name (especially if we want to divide the code into lib + scripts) might need additional changes/elaboration.

PR #21, opened 2020-10-30
Rework simple capture

vogelpi:
This PR slightly reworks how the simple capture scripts handles failures (now retries in case of failures and prints a warning, previously it just silently skipped failed captures). Also, a configurable delay is added before the capture start in order to unplug the FTDI cable to reduce noise.

### DIFF
--- a/cw/cw305/simple_capture_traces.py
+++ b/cw/cw305/simple_capture_traces.py
@@ -56,23 +56,36 @@ def run_capture(capture_cfg, ot, ktp):
   project = cw.create_project(project_file, overwrite=True)
 
   for i in tqdm(range(capture_cfg['num_traces']), desc='Capturing', ncols=80):
-    key, text = ktp.next()
-    ret = cw.capture_trace(ot.scope, ot.target, text, key)
-    if not ret:
-      print('Failed capture')
-      continue
-    # This value may need to be updated if the trace dB factor changes.
-    elif min(ret.wave) < -0.45:
-      continue
+    done = False
+    # Get at most one print about failures per status bar percent.
+    if i % (int(capture_cfg['num_traces']/100)) == 0:
+      print_fail = True
+    # The capture might fail, we keep trying until we succeed.
+    while not done:
+      key, text = ktp.next()
+      ret = cw.capture_trace(ot.scope, ot.target, text, key)
+      if not ret:
+        if print_fail:
+          print(f' Capture failed. Keep trying.')
+          print_fail = False
+        continue
+      # This value may need to be updated if the trace dB factor changes.
+      elif min(ret.wave) < -0.45:
+        if print_fail:
+          print(f' Signal too strong. Discard capture, keep trying.')
+          print_fail = False
+        continue
+      else:
+        done = True
 
-    expected = binascii.b2a_hex(cipher.encrypt(bytes(text)))
-    got = binascii.b2a_hex(ret.textout)
-    assert (got == expected), (
-        f'Incorrect encryption result!\ngot: {got}\nexpected: {expected}\n')
+      expected = binascii.b2a_hex(cipher.encrypt(bytes(text)))
+      got = binascii.b2a_hex(ret.textout)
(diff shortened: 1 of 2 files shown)


cw/cw305/simple_capture_traces.py
```diff
@@ -56,23 +56,36 @@ def run_capture(capture_cfg, ot, ktp):
...
+    while not done:
+      key, text = ktp.next()
+      ret = cw.capture_trace(ot.scope, ot.target, text, key)
+      if not ret:
+        if print_fail:
+          print(f' Capture failed. Keep trying.')
+          print_fail = False
+        continue
+      # This value may need to be updated if the trace dB factor changes.
+      elif min(ret.wave) < -0.45:
```
vogelpi, line 73: @moidx should we document/motivate this particular value?

moidx: I think this was just a workaround. Feel free to remove it. We should be able to remove these spurious traces with post processing.

cw/cw305/simple_capture_traces.py
```diff
@@ -56,23 +56,36 @@ def run_capture(capture_cfg, ot, ktp):
...
-
-    project.traces.append(ret)
-
-    ot.target.flush()
+    done = False
+    # Get at most one print about failures per status bar percent.
+    if i % (int(capture_cfg['num_traces']/100)) == 0:
+      print_fail = True
+    # The capture might fail, we keep trying until we succeed.
+    while not done:
```
moidx, line 64: I'd rather keep the code simple instead of implementing a nested loop. What is the motivation behind trying indefinitely instead of increasing 'num_traces'?

moidx (commented):
Hi @vogelpi, I added a couple of comments.

Thanks

cw/cw305/simple_capture_traces.py
vogelpi: Thanks @moidx for your review.

The motivation is this: Previously, we were just dropping such traces without informing the user. As a result, it was possible that with wrong gain settings, the script would trigger potentially hundreds of thousands of encryptions without recording a single trace. After hours of waiting, the user might end up with an empty trace file.

The least we should do is to print a warning that the trace was dropped (as we do in case communication fails and `ret` is 0).

Trying to repeat the capture in case the signal is to strong is motivated by the fact that the signal strength depends on the data processed (data input, key, masks). Depending on the gain settings and randomness, it may be possible to properly record 10000 traces first and then see a couple of failures. My intention was that if we specify to collect 100000 traces, we should indeed record 100000 traces. This is where the nested loop comes from.

Some additional complexity is added by the printing. `tqdm` seems not to support messages to be printed by the loop itself and my idea was that the user should not be spammed by warning messages. So per status percent point, we get at most warning message.

cw/cw305/simple_capture_traces.py
vogelpi: Ah, I see. So this is just about spurious noise. I thought that we wanted to limit the signal strength and avoid saturation in the ADC.

cw/cw305/simple_capture_traces.py
moidx: Thanks for elaborating on the rationale. I think running an initial capture with 10-100 traces to verify the setup should be sufficient. 

After removing the amplitude threshold filter workaround you will always get the number of traces requested. With a stable setup I haven't seen the number of traces drop significantly after post processing (i.e. dropping bad traces), but there is always a number of traces being scrapped ~5%, so I wouldn't add complexity just to get to the number of traces. 

I don't think the issue with the signal strength for a given input data should be a problem here either. If there is such a dependency, then that would be better handled with a custom key/textin generator.

vogelpi: Thanks @moidx for your input. I totally agree with you, the signal strength issue is better addressed differently. We just let the user know if the scope gain should be reduced after an initial capture run. I have filed a new PR #22 to achieve this and will close this one.

PR #22, opened 2020-11-03
Rework simple capture

vogelpi:
This PR slightly reworks how the simple capture scripts handles cases where the ADC might saturate. Previously it just silently skipped failed captures. Now, the trace is recorded anyway. But when plotting the recorded traces, all traces are checked for samples outside a safe range and a warning is printed in that case. The user is suggested to reduce the scope gain. This allows to keep the capture script simple but gives an easy way to verify the setup before triggering a long capture run.

Also, a configurable delay is added before the capture start in order to unplug the FTDI cable to reduce noise.

This PR supersedes #21 .

### DIFF
--- a/cw/cw305/simple_capture_traces.py
+++ b/cw/cw305/simple_capture_traces.py
@@ -80,6 +80,11 @@ def run_capture(capture_cfg, ot, ktp):
 def plot_results(plot_cfg, project_name):
   """Plots traces from `project_name` using `plot_cfg` settings."""
   project = cw.open_project(project_name)
+
+  if len(project.waves) == 0:
+    print('Project contains no traces. Did the capture fail?')
+    return
+
   plot.save_plot_to_file(project.waves, plot_cfg['num_traces'],
                          plot_cfg['trace_image_filename'])
 
--- a/cw/cw305/simple_capture_traces.py
+++ b/cw/cw305/simple_capture_traces.py
@@ -61,9 +61,6 @@ def run_capture(capture_cfg, ot, ktp):
     if not ret:
       print('Failed capture')
       continue
-    # This value may need to be updated if the trace dB factor changes.
-    elif min(ret.wave) < -0.45:
-      continue
 
     expected = binascii.b2a_hex(cipher.encrypt(bytes(text)))
     got = binascii.b2a_hex(ret.textout)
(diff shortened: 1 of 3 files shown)


cw/cw305/simple_capture_traces.py
```diff
@@ -61,9 +61,6 @@ def run_capture(capture_cfg, ot, ktp):
     if not ret:
       print('Failed capture')
       continue
-    # This value may need to be updated if the trace dB factor changes.
-    elif min(ret.wave) < -0.45:
-      continue
```
alphan, line 66: I also removed this part in #23 but I will drop that commit since this PR is the right place for this change.

cw/cw305/simple_capture_traces.py
```diff
@@ -80,6 +77,21 @@ def run_capture(capture_cfg, ot, ktp):
...
+
+  # The ADC output is in the interval [-0.5, 0.5). Check that the recorded
+  # traces are within that range with some safety margin.
+  if not (np.all(np.greater(project.waves, -plot_cfg['amplitude_max'])) and
+          np.all(np.less(project.waves, plot_cfg['amplitude_max']))):
+    print('WARNING: Some traces have samples outside the range (' +
+          str(-plot_cfg['amplitude_max']) + ', ' +
+          str(plot_cfg['amplitude_max']) + ').')
+    print('\t The ADC has a max range of [-0.5, 0.5) and might saturate.')
+    print('\t It is recommended to reduce the scope gain (see device.py).')
```
alphan, line 93: nit: Can you replace `\t` with spaces so that these lines are aligned with the text after `WARNING: `?

cw/cw305/simple_capture_traces.py
```diff
@@ -95,6 +107,10 @@ def parse_args():
                       '-p',
                       type=int,
                       help="Plot number of traces.")
+  parser.add_argument('--start-delay',
```
alphan, line 110: I think you can drop changes related to start delay since we will be using SAM3U for programming.

alphan (approved):
Thanks @vogelpi! I think you can drop the changes related to start-delay since we will be using SAM3U but otherwise LGTM. Thanks for adding those warnings, it took me a while to figure out what was going on when this happened on my setup :)

moidx (approved):
LGTM!

cw/cw305/simple_capture_traces.py
vogelpi: Done.

cw/cw305/simple_capture_traces.py
vogelpi: Yep, I agree. Dropped the corresponding commit.

cw/cw305/simple_capture_traces.py
vogelpi: Thanks.

vogelpi:
Thanks guys for your review. I have updated the PR and will now merge it.

PR #25, opened 2020-11-06
Minor updates to cw_to_trs.py

alphan:
This change adds a progress bar and an informational message to indicate
the status of the conversion process. This change also modifies the
`--export-key` argument so that it doesn't expect a `True` or `False`
value.


moidx (approved):
LGTM

PR #28, opened 2020-12-16
Add a distributed implementation of the correlation-enhanced collision attack

alphan:
This PR adds a distributed implementation of the correlation-enhanced power analysis collision attack described in "Correlation-Enhanced Power Analysis Collision Attack" by A. Moradi, O. Mischke, and T. Eisenbarth (https://eprint.iacr.org/2010/297.pdf).

In addition to using Ray (https://docs.ray.io/en/master/index.html) for distributed computation, this implementation also uses a modified version of Dijkstra's shortest path algorithm to utilize all available information on differences between key bytes. This implementation also optimizes several operations for improved performance, e.g. various `numpy` operations and encrypting with multiple keys in parallel using `scared` to check if the attack was successful, etc.

I get the following numbers when I run this on my workstation (12 cores @ 3.7 GHz according to `/proc/cpuinfo`) using the traces in `aes_unhardened_1mio_traces_cw305.tar.bz2` captured by @vogelpi:
```
> ./ceca.py -f projects/opentitan_simple_aes.cwp -n 400000 -w 10 -a 117 127 -d output -s 3

2020-12-16 09:18:45,414 INFO services.py:1090 -- View the Ray dashboard at http://127.0.0.1:8265
2020-12-16 09:18:48,020 INFO ceca.py:489 -- Will use 397502 traces (99.4% of all traces)
2020-12-16 09:18:48,586 INFO _timer.py:57 -- compute_pairwise_diffs_and_scores took 0.4s
2020-12-16 09:18:48,588 INFO ceca.py:497 -- Difference values (delta_0_i): [  0 196  41 120  25  62 245  89  49 239 220  24 102 179 220 118]
2020-12-16 09:18:48,669 INFO ceca.py:502 -- Recovered AES key: 2b7e151628aed2a6abf7158809cf4f3c
2020-12-16 09:18:48,670 INFO _timer.py:57 -- perform_attack took 1.8s
2020-12-16 09:18:48,670 INFO _timer.py:57 -- main took 3.8s
```

```
> ./ceca.py -f projects/opentitan_simple_aes.cwp -n 1000000 -w 10 -a 110 130 -d output -s 3

2020-12-16 09:19:12,316 INFO services.py:1090 -- View the Ray dashboard at http://127.0.0.1:8265
2020-12-16 09:19:15,437 INFO ceca.py:489 -- Will use 989639 traces (99.0% of all traces)
2020-12-16 09:19:16,181 INFO _timer.py:57 -- compute_pairwise_diffs_and_scores took 0.4s
2020-12-16 09:19:16,183 INFO ceca.py:497 -- Difference values (delta_0_i): [  0 196  41 120  25  62 245  89  49 239 220  24 102 179 220 118]
2020-12-16 09:19:16,264 INFO ceca.py:502 -- Recovered AES key: 2b7e151628aed2a6abf7158809cf4f3c
2020-12-16 09:19:16,265 INFO _timer.py:57 -- perform_attack took 2.5s
2020-12-16 09:19:16,265 INFO _timer.py:57 -- main took 4.4s
```

The difference between `main()` and `perform_attack()` is Ray's initialization overhead.

I am aware that some of the code can be moved to utility libraries, I plan to revisit those after running this on GCP.

Looking forward to your feedback!

### DIFF
--- a/cw/cw305/ceca.py
+++ b/cw/cw305/ceca.py
@@ -1,0 +1,591 @@
+#!/usr/bin/env python3
+# Copyright lowRISC contributors.
+# Licensed under the Apache License, Version 2.0, see LICENSE for details.
+# SPDX-License-Identifier: Apache-2.0
+
+import argparse
+import enum
+import logging
+import sys
+
+import chipwhisperer as cw
+import chipwhisperer.analyzer as cwa
+import codetiming
+import more_itertools
+import networkx as nx
+import numpy as np
+import ray
+import scared
+
+"""A distributed implementation of the correlation-enhanced power analysis
+collision attack.
+
+See "Correlation-Enhanced Power Analysis Collision Attack" by A. Moradi, O.
+Mischke, and T. Eisenbarth (https://eprint.iacr.org/2010/297.pdf) for more
+information.
+
+Typical usage:
+>>> ./ceca.py -f PROJECT_FILE -n 400000 -w 5 -a 117 127 -d output -s 3
+"""
+
+
+def timer():
+    """A customization of the ``codetiming.Timer`` decorator."""
+
+    def decorator(func):
+        @codetiming.Timer(
+            name=func.__name__,
+            text=f"{func.__name__} took {{seconds:.1f}}s",
+            logger=logging.info,
+        )
+        def wrapper(*args, **kwargs):
+            return func(*args, **kwargs)
+
+        return wrapper
+
+    return decorator
+
+
+class AttackDirection(str, enum.Enum):
+    """Enumeration for attack direction."""
+
+    INPUT = "input"
+    OUTPUT = "output"
+
+
+@ray.remote
+class TraceWorker:
+    """Class for performing distributed computations on power traces.
+
+    This class provides methods for performing distributed computations on power
+    traces such as computing the mean, the standard deviation, and filtering.
+    After creating multiple instances (workers) of this class and initializing
+    each worker with a subset of the available traces, distributed computations
+    can be performed by simply calling the methods of these workers. Individual
+    results of these workers can then be aggregated to produce the final result.
+
+    This class is a Ray actor (https://docs.ray.io/en/master/index.html) and can
+    be used as follows:
+    >>> workers = [TraceWorker.remote(...) for ...]
+    >>> tasks = [worker.compute_stats.remote() for worker in workers]
+    >>> results = ray.get(tasks)
+    """
+
+    def __init__(self, project_file, trace_slice, attack_window, attack_direction):
+        """Inits a TraceWorker.
+
+        Args:
+            project_file: A Chipwhisperer project file.
+            trace_slice: Traces assigned to this worker.
+            attack_window: Samples to process.
+            attack_direction: Attack direction.
+        """
+        self.project = cw.open_project(project_file)
+        # TODO: Consider more efficient formats.
+        trace_cnt = trace_slice.stop - trace_slice.start
+        self.num_samples = attack_window.stop - attack_window.start
+        if attack_direction == AttackDirection.INPUT:
+            self.texts = np.vstack(self.project.textins[trace_slice])
+        else:
+            self.texts = np.vstack(self.project.textouts[trace_slice])
+        self.traces = np.asarray(self.project.waves[trace_slice])[:, attack_window]
+
+    def compute_stats(self):
+        """Computes sums and sums of deviation products of traces.
+
+        Results from multiple workers can be aggregated to compute the standard
+        deviation of a set of traces in a distributed manner using Eq. 22 in
+        "Numerically Stable Parallel Computation of (Co-)Variance" by E.
+        Schubert and M. Gertz (https://dbs.ifi.uni-heidelberg.de/files/Team/
+        eschubert/publications/SSDBM18-covariance-authorcopy.pdf).
+
+        Returns:
+            Number of traces, their sums, and sums of deviation products.
+        """
+        cnt = self.traces.shape[0]
+        sum_ = self.traces.sum(axis=0)
+        mean = sum_ / cnt
+        sum_dev_prods = ((self.traces - mean) ** 2).sum(axis=0)
+        return (cnt, sum_, sum_dev_prods)
+
+    def filter_noisy_traces(self, min_trace, max_trace):
+        """Filters traces with values outside the allowable range.
+
+        Args:
+            min_trace: Minimum allowable values.
+            max_trace: Maximum allowable values.
+
+        Returns:
+            Number of remaining traces.
+        """
+        traces_to_use = np.all(
+            (self.traces >= min_trace) & (self.traces <= max_trace), axis=1
+        )
+        self.traces = self.traces[traces_to_use]
+        self.texts = self.texts[traces_to_use]
+        return self.traces.shape[0]
+
+    def count_and_sum_text_traces(self):
+        """Computes the number of traces and sums of these traces for all values
+        of each text byte.
+
+        Returns:
+            A tuple ``(cnts, sums)``, where
+                - ``cnts`` is a (16, 256, 1) array where ``cnts[i, j, 0]`` gives the
+                  number of traces where text byte i is j, and
+                - ``sums`` is a (16, 256, NUM_SAMPLES) array where ``sums[i, j, :]``
+                  gives the sum of traces where text byte i is j.
+        """
+        sums = np.zeros((16, 256, self.num_samples))
+        # Need to specify the last dimension for broadcasting to work during
+        # aggregation.
+        cnts = np.zeros((16, 256, 1))
+        for byte_pos in range(16):
+            # While a little bit more complex, below code is more efficient than
+            # a naive implementation that searches for all possible byte values
+            # in ``self.texts``.
+            sorted_indices = self.texts[:, byte_pos].argsort()
+            sorted_bytes = self.texts[sorted_indices, byte_pos]
+            # Find the indices where byte values change.
+            val_changes = np.where(np.roll(sorted_bytes, 1) != sorted_bytes)[0]
+            # Append the number of rows to be able to use ``pairwise``.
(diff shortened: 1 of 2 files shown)


cw/cw305/ceca.py
```diff
@@ -0,0 +1,574 @@
...
+    # array where betas[i, j] = diffs[i] ^ alphas[j]
+    betas = alphas ^ diffs
+    # Compute the most likely differences between all pairs of plaintext bytes.
+    for a in range(16):
+        for b in range(a + 1, 16):
+            corrcoefs = np.corrcoef(mean_traces[a], mean_traces[b])
+            # np.corrcoef returns the values that we want in the upper right
+            # quadrant.
+            diff_corrcoefs = corrcoefs[alphas, 256 + betas].sum(axis=1)
+            best_diff = diff_corrcoefs.argmax()
```
vogelpi, line 284: It's not fully clear how your implementation works but you basically managed to skip three more loops (compared to mine) which is great :+1:

vogelpi: What I am wondering: you don't seem to normalize the correlation coefficients BEFORE doing the argmax. If I remember correctly, this was necessary in my implementation. But still yours works correctly....

cw/cw305/ceca.py
```diff
@@ -0,0 +1,574 @@
...
+            # TODO: Analyze the effect of /diff_corrcoefs.mean() below.
+            pairwise_diffs_scores[(a, b), (b, a)] = (best_diff,
+                                                     diff_corrcoefs[best_diff] /
+                                                     diff_corrcoefs.mean())
```
vogelpi, line 288: Ah, here comes the normalization. I think you need this to make all the `pairwise_diffs_scores` comparable. There are pairs for which the correlation is almost flat but constantly higher than the highest peaks of another pair. You need the normalization to "downvote" this flat correlation for `find_best_diffs()`.

vogelpi (approved):
Thanks @alphan for this PR, it's awesome work!

LGTM, I just have a couple of minor comments.

As an extension in a follow up PR, do you think it would be possible to implement a sweep over the number of traces? This would provide the useful information of when the attack starts to succeed, i.e., what the resistance in terms of number of traces actually is. Anyway, worst case we can just wrap your attack with a sweep script. Yours is so incredibly fast now that this would still be okay I guess.

cw/cw305/ceca.py
```diff
@@ -0,0 +1,574 @@
...
+        cnt, sum_, sum_dev_prods = ray.get(done[0])
+        if running_cnt == 0:
+            running_sum_dev_prods = np.copy(sum_dev_prods)
+            running_sum = np.copy(sum_)
+            running_cnt += cnt
+        else:
+            running_sum_dev_prods += sum_dev_prods + (cnt * running_sum -
+                                                      running_cnt * sum_)**2 / (
+                                                          cnt * running_cnt *
+                                                          (cnt + running_cnt))
```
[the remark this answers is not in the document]
moidx: +1, you can consider adding everything inside parenthesis to get more flexibility on alignment if needed.

python-requirements.txt
```diff
@@ -4,11 +4,17 @@
...
+more-itertools
+networkx
 pycryptodome
 pyyaml
+ray
 scared
 tqdm
 trsfile
+# can be removed after switching to ray
+joblib
```
moidx, line 17: Is this still needed?

moidx (approved):
LGTM, great work!

cw/cw305/ceca.py
```diff
@@ -0,0 +1,574 @@
...
+import scared
+"""A distributed implementation of the correlation-enhanced power analysis
+collision attack.
+
+See "Correlation-Enhanced Power Analysis Collision Attack" by A. Moradi, O.
+Mischke, and T. Eisenbarth (https://eprint.iacr.org/2010/297.pdf) for more
+information.
+
+Typical usage:
+>>> ./ceca.py -f PROJECT_FILE -n 400000 -w 5 -a 117 127 -d output -s 3
```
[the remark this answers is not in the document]
alphan: Oops, this was because arguments were not marked as required. Fixed.

cw/cw305/ceca.py
```diff
@@ -0,0 +1,574 @@
...
+        Returns:
+            Number of remaining traces.
+        """
+        traces_to_use = np.all(
+            (self.traces >= min_trace) & (self.traces <= max_trace), axis=1)
+        self.traces = self.traces[traces_to_use]
+        self.texts = self.texts[traces_to_use]
+        return self.traces.shape[0]
+
+    def count_and_sum_plaintext_traces(self):
```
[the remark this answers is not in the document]
alphan: Good point, I implemented the input direction first, will fix all such occurrences.

cw/cw305/ceca.py
```diff
@@ -0,0 +1,574 @@
...
+        """Computes the number of traces and sums of these traces for all values
+        of each plaintext byte.
```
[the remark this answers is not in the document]
alphan: Edited the "Returns" section to include this information.

cw/cw305/ceca.py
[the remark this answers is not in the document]
alphan: I will reformat the code with `black`.

cw/cw305/ceca.py
```diff
@@ -0,0 +1,574 @@
...
+        mean_plaintext_traces)
+    diffs = find_best_diffs(pairwise_diffs_scores)
+    logging.info(f'Difference values (delta_0_i): {diffs}')
+    # Recover the key.
+    key = recover_key(diffs, attack_direction, project.textins[0],
+                      project.textouts[0])
+    if key is not None:
+        logging.info(f'Recovered AES key: {bytes(key).hex()}')
+    else:
+        logging.error(f'Failed to recover the AES key')
```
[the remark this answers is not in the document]
alphan: I think so, but based on my limited experiments this sometimes can be a red herring because of the noise and the way we utilize difference values. That's why I didn't include it in this PR (also to keep things simple) but I don't think it will introduce a noticeable overhead and should be easy to add. Created #30.

cw/cw305/ceca.py
```diff
@@ -0,0 +1,574 @@
...
+See "Correlation-Enhanced Power Analysis Collision Attack" by A. Moradi, O.
+Mischke, and T. Eisenbarth (https://eprint.iacr.org/2010/297.pdf) for more
+information.
+
+Typical usage:
+>>> ./ceca.py -f PROJECT_FILE -n 400000 -w 5 -a 117 127 -d output -s 3
+"""
+
+
+def timer():
```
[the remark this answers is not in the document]
alphan: I would prefer 4 spaces to align with PEP8 and leave these to automated tools. I think `black` is a good choice.

python-requirements.txt
alphan: Yes, this is required by [this line](https://github.com/lowRISC/ot-sca/blob/8e61b89075ae5a886d35090b2675d721705d495c/cw/cw305/correlation-enhanced_collision_attack.py#L18) in `correlation-enhanced_collision_attack.py`.

cw/cw305/ceca.py
alphan: I think so too but I left a TODO nevertheless because IIRC I was able to recover the key without this a couple of times and would like to check how the results look like when running batch attacks.

cw/cw305/ceca.py
alphan: > What I am wondering: you don't seem to normalize the correlation coefficients BEFORE doing the argmax. If I remember correctly, this was necessary in my implementation. But still yours works correctly....

IIUC, you are referring to this snippet (lines [98](https://github.com/lowRISC/ot-sca/blob/8e61b89075ae5a886d35090b2675d721705d495c/cw/cw305/correlation-enhanced_collision_attack.py#L98)-104 in `correlation-enhanced_collision_attack.py`):
```python
    for i in range(256):
        for j in range(256):
            delta = i ^ j
            rho_avg_delta[delta] += rho_mat[i, j]
            rho_avg_delta_num[delta] += 1
    for delta in range(256):                                # redundant
        rho_avg_delta[delta] /= rho_avg_delta_num[delta]    # redundant
    # Normalize correlations.
    mean = rho_avg_delta.mean()
    rho_avg_delta /= mean
```

Unless I'm missing something, it looks like the loop marked by `redundant` doesn't really do anything, maybe except for introducing some tiny numerical error due to floating point operations. Since `rho_avg_delta_num[delta]` is `256` for all values of `delta`, i.e. [0, 255] and `rho_avg_delta` is normalized again, this loop is basically dividing all entries by some constant before they are normalized.

alphan: > It's not fully clear how your implementation works but you basically managed to skip three more loops (compared to mine) which is great 👍

Thanks :) The main point here is to minimize the back and forth between python and numpy as much as possible, e.g. by avoiding unnecessary (nested) loops that can be handled by numpy much more efficiently. In a nutshell:
- `alphas` is a (256,) row array, `diffs` is a (256, 1) col array (so that broadcasting works out), and `betas` is a (256, 256) array where `betas[i,j] = diffs[i] ^ alphas[j]`.
- Defining `betas` this way is important because this lets us pick the values we want from the `np.corrcoeff` result using just a single indexing operation and a `sum(axis=1)` as opposed to two `range(0, 256)` nested loops.

I hope this helps and please let me know if you have any other questions.

alphan:
Thanks for your reviews @vogelpi and @moidx!
In the last push:
* Addressed comments
* Reformatted code with `black`.
* Fixed some minor typos in help strings.

cw/cw305/ceca.py
vogelpi: Thanks for the clarification @alphan .

Regarding the redundant loop above: I added this (before doing the real normalization) because I wasn't sure if there are equally many combinations of `i` and `j` that result in the same `delta`. But yes, I think with the real normalization in place this loop can actually be dropped.

PR #29, opened 2021-01-07
Allow changing #samples used for correlation-enhanced collision attack

vogelpi:
The number of samples used for computing the average traces was erroneously hardcoded.


moidx (approved):
LGTM

vogelpi:
Thanks @moidx !

PR #31, opened 2021-01-19
Add crosshair and hover tools to bokeh html output

alphan:
This change adds crosshair and hover tools to bokeh html output:

image


vogelpi (approved):
That's useful, thanks @alphan . LGTM!

PR #35, opened 2021-02-12
Add WaveRunner support and batch capture script

alphan:
This change adds support for capturing traces using a LeCroy WaveRunner 9104 and a new script that captures traces using the new 'b' (batch encrypt) command.

As I mentioned earlier, WaveRunner has several settings that we should evaluate. For the sake of simplicity, most of the parameters are hardcoded for now in `waverunner.py`. I thought it would be better to postpone the decision on which parameters should be fixed and which should be tunable when we start working on the CLI (and the refactoring that will go with it). I guess the most important parameters are sampling rate and noise filtering which are set to 1 GS/s down to 100 MS/s with 10x sparsing + 2.5bits ERes (enhanced resolution, i.e. noise filtering). 2.5bits ERes is a bit extreme and significantly reduces the bandwidth (-3dB @ 14.50 MHz according to the oscilloscope) but it performed better than smaller values in the ceca attack so I decided to leave it as-is. This obviously results in much smoother traces when compared to chipwhisperer (top: waverunner, bottom: chipwhisperer):

image

image

I confirmed that these changes work using an older `aes_serial` binary and the bitstream in the ot-sca repo (will try building these from opentitan:master next). This change lets us capture traces at a rate of ~1200 traces/s:
```
$ ./simple_capture_traces_waverunner.py
Connecting and loading FPGA
Initializing PLL1
Programming OpenTitan with "objs/aes_serial_fpga_nexysvideo.bin"...
Transferring frame 0x00000000 @ 0x00000000.
...
Transferring frame 0x8000000B @ 0x00002A48.
Serial baud rate = 38400
Serial baud rate = 115200
Scope setup with sampling rate 100003063.0 S/s
Target simpleserial version: 'z01'.
Using key: '2b7e151628aed2a6abf7158809cf4f3c'.
Connected to LECROY WAVERUNNER9104 (ip: 192.168.1.228, serial: LCRY4701N40366, version: 9.1.0, options: HDTV, JTA2, XDEV, XWEB)
Capturing: 100%|███████████████| 1000000/1000000 [13:18<00:00, 1252.54 traces/s]
```

I also confirmed that we can recover the key with the new traces:
```
$ ./ceca.py -f projects/opentitan_simple_aes -n 1000000 -w 8 -a 110 130 -d output -s 3
2021-02-11 21:04:07,420 INFO services.py:1171 -- View the Ray dashboard at http://127.0.0.1:8265
2021-02-11 21:04:10,909 INFO ceca.py:492 -- Will use 982919 traces (98.3% of all traces)
2021-02-11 21:04:11,656 INFO _timer.py:57 -- compute_pairwise_diffs_and_scores took 0.3s
2021-02-11 21:04:11,657 INFO ceca.py:501 -- Difference values (delta_0_i): [  0 196  41 120  25  62 245  89  49 239 220  24 102 179 220 118]
2021-02-11 21:04:11,734 INFO ceca.py:505 -- Recovered AES key: 2b7e151628aed2a6abf7158809cf4f3c
2021-02-11 21:04:11,735 INFO _timer.py:57 -- perform_attack took 2.9s
2021-02-11 21:04:11,735 INFO _timer.py:57 -- main took 4.9s
```

Initially I was planning to refactor the current capture script, but I thought it would be better to just add a new script without touching the one that everyone relies on. Again, something to tackle when we start working on the CLI.

### DIFF
--- a/cw/cw305/waverunner.py
+++ b/cw/cw305/waverunner.py
@@ -1,0 +1,250 @@
+# Copyright lowRISC contributors.
+# Licensed under the Apache License, Version 2.0, see LICENSE for details.
+# SPDX-License-Identifier: Apache-2.0
+
+"""Support for capturing traces using LeCroy WaveRunner 9104."""
+
+import logging
+import re
+
+import numpy as np
+import vxi11
+
+
+class _Timeout:
+    """Helper class for setting scoped timeout values."""
+
+    def __init__(self, instr, timeout):
+        self._instr = instr
+        self._orig_timeout = self._instr.timeout
+        self._instr.timeout = timeout
+
+    def __enter__(self):
+        return self
+
+    def __exit__(self, exc_type, exc_val, exc_tb):
+        self._instr.timeout = self._orig_timeout
+
+
+class WaveRunner:
+    """Class for capturing traces using a WaveRunner oscilloscope.
+
+    This class operates the oscilloscope in sequence mode to improve performance and in
+    2-channel mode to maximize sampling rate. Trigger and power signals must be
+    connected to channels 2 and 3, respectively.
+
+    When in sequence mode, the oscilloscope captures a total of `seq_num_waves` waves
+    each starting at a trigger event. This is much more efficient than sending a
+    separate command for each wave.
+
+    This class is only tested to work with WaveRunner 9104 series.
+
+    For more information on the commands used in this module please see:
+    - Operator's Manual WaveRunner 9000 and WaveRunner 8000-R Oscilloscopes
+        (http://cdn.teledynelecroy.com/files/manuals/waverunner-9000-operators-manual.pdf)
+    - MAUI Oscilloscopes Remote Control and Automation Manual
+        (http://cdn.teledynelecroy.com/files/manuals/maui-remote-control-and-automation-manual.pdf)
+
+    Typical usage:
+    >>> waverunner = WaveRunner('192.168.1.227')
+    >>> waverunner.configure()
+    >>> while foo:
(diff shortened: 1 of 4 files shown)


cw/cw305/waverunner.py
```diff
@@ -0,0 +1,246 @@
...
+    >>> waverunner.configure()
+    >>> while foo:
+    >>>     ...
+    >>>     waverunner.seq_num_waves = seq_num_waves
+    >>>     waverunner.arm()
+    >>>     waves = waverunner.wait_for_acquisition_and_transfer_waves()
+    >>>     ...
+
+    Attributes:
+        seq_num_waves: Number of waves per sequence.
```
moidx, line 55: nit: Can you elaborate a bit more about this? Is each wave equivalent to num_samples, and the scope just does seq_num_waves * num_samples after the trigger?

moidx (approved):
LGTM. Thanks for the detailed write up!

cw/cw305/waverunner.py
alphan: Each wave has num_samples sample, we send a single command to the oscilloscope to put in sequence mode (instead of talking to it for each wave) and it captures a total of seq_num_waves waves, each starting with a trigger. I've seen this being called segmented traces, repeated trigger, etc.

alphan: Updated class docstring.

alphan:
@vogelpi merging this now as we discussed since I need to make a few changes for new binaries. Please feel free to comment here or on a new issue. Thanks!

PR #36, opened 2021-02-13
Patch fpga.FPGAProgram at runtime to detect if the FPGA was actually programmed

alphan:
We currently don't force reprogramming the FPGA on CW305 if it's already programmed. This is fine except when you have a bitstream and forget about this behavior (like me). This change prints a more accurate message indicating whether the FPGA was actually programmed or not.

### DIFF
--- a/cw/cw305/util/device.py
+++ b/cw/cw305/util/device.py
@@ -4,10 +4,37 @@
 
 r"""CW305 utility functions. Used to configure FPGA with OpenTitan design."""
 
+import inspect
 import subprocess
 import time
 
 import chipwhisperer as cw
+
+class RuntimePatchFPGAProgram:
+    """Replaces the FPGAProgram method of an FPGA object with a function
+    that calls the given callback before calling the original method.
+
+    This class can be used to detect if the FPGA was actually programmed or not.
+    """
+    def __init__(self, fpga, callback):
+        """Inits a RuntimePatchFPGAProgram.
+
+        Args:
+            fpga: An FPGA object.
+            callback: Callback to call when fpga.FPGAProgram() is called.
+        """
+        self._fpga = fpga
+        self._callback = callback
+        self._orig_fn = fpga.FPGAProgram
+
+    def __enter__(self):
+        def wrapped_fn(*args, **kwargs):
+            self._callback()
+            return self._orig_fn(*args, **kwargs)
+        self._fpga.FPGAProgram = wrapped_fn
+
+    def __exit__(self, exc_type, exc_value, traceback):
+        self._fpga.FPGAProgram = self._orig_fn
 
 class OpenTitan(object):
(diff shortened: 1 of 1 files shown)


cw/cw305/util/device.py
```diff
@@ -17,10 +44,29 @@ def __init__(self, fw_programmer, bitstream, pll_frequency, baudrate):
...
     # Do not program the FPGA if it is already programmed.
-    fpga.con(bsfile=bitstream, force=False)
+    # Note: Set this to True to force programming the FPGA when using a new
+    # bitstream.
+    # TODO: We should have this in the CLI.
+    force_programming = False
+    print('Connecting and loading FPGA... ', end = '')
+    fpga = cw.capture.targets.CW305()
+    # Runtime patch fpga.fpga.FPGAProgram to detect if it was actually called.
+    # Note: This is fragile and may break but it is easy to miss that the FPGA
```
moidx, line 55: Maybe we can add a hard reset command to always have a deterministic way of initializing the FPGA (whenever needed).

moidx (approved):
LGTM

I do agree, this solution seems fragile. Maybe we should look at pushing a patch to CW to determine if an update is going to take effect, or something else?

cw/cw305/util/device.py
alphan: Yes, we should definitely expose this as a command line argument. But still, in order to print an accurate message (or to program the FPGA only if needed) we need to know the last bitstream that the FPGA was programmed with (chipwhisperer doesn't know this either AFAICT, it just checks a flag in a status packet). We can try writing the last programmed hash to a file but that would require us to introduce an on-disk cache. If we decide to do that I think this should definitely go there.

moidx: I agree, this is an orthogonal comment. If one wants to make sure the new FPGA is loaded it can force a hard reset before running.

alphan:
Thanks for the quick review @moidx.

> I do agree, this solution seems fragile. Maybe we should look at pushing a patch to CW to determine if an update is going to take effect, or something else?

Yes, we should try doing that. I took a quick look at `_con` in `CW305.py` and it looks like the code both connects to the FPGA and programs it there so we don't really have a way to check this before programming and after programming the FPGA is already programmed. Since the function doesn't return anything now, we could ask NewAE to consider returning something to indicate if the FPGA was actually programmed or not.

alphan:
I took another look at CW305.py and noticed that there is `get_fpga_buildtime` but unfortunately it doesn't seem to work properly now. Created newaetech/chipwhisperer#306 and newaetech/chipwhisperer#307.

vogelpi (approved):
Thanks @alphan for checking this out (and for triggering the other discussions). This may be a bit fragile but it's definitely suitable until we have the option working that can read the build time of the programmed bitstream.

Issue #24, opened 2020-11-05
aes_serial does not send a response for the first check version command


alphan:
Noticed this while working on #23 (should not be related). When I run `simple_capture_traces.py`. If I send multiple commands, it sends a response for each one except for the first one.

alphan:
Also noticed this when working on lowRISC/opentitan#4972: It looks like we are having problems receiving the first command. Occasionally it gets corrupted and for everything to work we need to ignore such commands. AFAICT this only happens for the first host -> device command. I didn't have a chance to look at this more closely but this could be because of some garbled characters ending up in the FIFO right after init or a clock issue.

johannheyszl:
closing this since the important aspects are covered in: https://github.com/lowRISC/ot-sca/issues/83

PR #38, opened 2021-02-13
Ping target with 'v' until it responds.

alphan:
This is related to #24 and doesn't fix the underlying issue (I'm still not sure what it is, could be the `flush` after setting the baudrate in `initialize_target` or something else) but I thought it would be a good idea to wait until receiving a proper response from the target before starting the capture.

### DIFF
--- a/cw/cw305/simple_capture_traces.py
+++ b/cw/cw305/simple_capture_traces.py
@@ -6,6 +6,7 @@ import argparse
 import binascii
 from Crypto.Cipher import AES
 import numpy as np
+import sys
 import time
 from tqdm import tqdm
 import yaml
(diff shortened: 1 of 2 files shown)


cw/cw305/simple_capture_traces.py
```diff
@@ -34,6 +35,19 @@ def initialize_capture(device_cfg, spiflash_cfg):
...
     device_cfg['baudrate'])
   print(f'Scope setup with sampling rate {ot.scope.clock.adc_rate} S/s')
+  # Ping target
+  print("Reading from FPGA using simpleserial protocol.")
+  version = None
+  ping_cnt = 0
+  while not version:
+    if ping_cnt == 3:
+      print(f"Error: No response from the target (attempts: {ping_cnt}).", file=sys.stderr)
+      sys.exit(1)
```
moidx, line 45: Should this throw an error instead, in case this is used in the context of a library?

cw/cw305/simple_capture_traces.py
```diff
@@ -34,6 +35,19 @@ def initialize_capture(device_cfg, spiflash_cfg):
     device_cfg['pll_frequency'],
     device_cfg['baudrate'])
   print(f'Scope setup with sampling rate {ot.scope.clock.adc_rate} S/s')
+  # Ping target
+  print("Reading from FPGA using simpleserial protocol.")
```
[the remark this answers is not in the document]
alphan: Good catch, fixed. I ran a formatter in a different file which happens to favor `"`. Style guide says both is fine but we may end up choosing `"` to be consistent with `"""`.

cw/cw305/simple_capture_traces.py
alphan: Good point, done.

moidx (approved):
LGTM

vogelpi (approved):
LGTM, thanks @alphan !

PR #37, opened 2021-02-13
Update FPGA bitstream and aes_serial binary

alphan:
This change updates the FPGA bitstream and `aes_serial` binary. Since I wasn't able to generate the bitstream from the tip (lowRISC/opentitan@445fe3132), both files were built by cherry-picking the relevant `aes_serial` updates (lowRISC/opentitan@675e9497b and lowRISC/opentitan@cbbc70fe2) on top of lowRISC/opentitan@71d98f8e (last known good commit).

This change also
  * Removes acks during capture (see cw/cw305/simple_capture_traces.py) and adjusts the scope gain in cw/cw305/util/device.py (not sure if the new value will be the best for everyone).
  * Adjusts the vertical scale and the trigger offset of waverunner.

Here's how the traces look (top: chipwhisperer, bottom: waverunner):

image

image

 I also verified that all capture scripts work with the new binaries and we can recover the key (tried only using waverunner):

```
$ ./simple_capture_traces_waverunner.py 
Connecting and loading FPGA
Initializing PLL1
Programming OpenTitan with "objs/aes_serial_fpga_nexysvideo.bin"...
Transferring frame 0x00000000 @ 0x00000000.
...
Transferring frame 0x8000000B @ 0x00002A48.
Serial baud rate = 38400
Serial baud rate = 115200
Scope setup with sampling rate 100003051.0 S/s
Reading from FPGA using simpleserial protocol.
Target simpleserial version: 'z01'.
Using key: '2b7e151628aed2a6abf7158809cf4f3c'.
Connected to LECROY WAVERUNNER9104 (ip: 192.168.1.228, serial: LCRY4701N40366, version: 9.1.0, options: HDTV, JTA2, XDEV, XWEB)
Capturing: 100%|███████████████| 2000000/2000000 [30:06<00:00, 1107.05 traces/s]
WARNING: Some traces have samples outside the range (-0.48, 0.48).
         The ADC has a max range of [-0.5, 0.5) and might saturate.
         It is recommended to reduce the scope gain (see device.py).

$ ./ceca.py -f projects/opentitan_simple_aes -n 1500000 -a 115 120 -w 8 -d output -s 3
2021-02-13 13:45:12,162 INFO services.py:1171 -- View the Ray dashboard at http://127.0.0.1:8265
2021-02-13 13:45:16,600 INFO ceca.py:492 -- Will use 1490514 traces (99.4% of all traces)
[...]

PR #40, opened 2021-02-16
[spiflash] Re-align frame size with OpenTitan, adjust wait times

vogelpi:
This PR is required to make ot-sca work again with current OpenTitan master. We should probably merge #37 first. WDYT @alphan ?

### DIFF
--- a/cw/cw305/util/spiflash.py
+++ b/cw/cw305/util/spiflash.py
@@ -60,7 +60,7 @@ class _SpiFlashFrame:
   DIGEST_SIZE = HASH_FUNCTION().digest_size
   FRAME_NUMBER_SIZE = 4
   FLASH_OFFSET_SIZE = 4
-  PAYLOAD_SIZE = 1024 - DIGEST_SIZE - FRAME_NUMBER_SIZE - FLASH_OFFSET_SIZE
+  PAYLOAD_SIZE = 2048 - DIGEST_SIZE - FRAME_NUMBER_SIZE - FLASH_OFFSET_SIZE
 
   def __init__(self, frame_number, flash_offset, payload):
     """Inits a _SpiFlashFrame.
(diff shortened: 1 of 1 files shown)


cw/cw305/util/spiflash.py
```diff
@@ -126,8 +126,8 @@ class _Bootstrap:
   _PIN_BOOTSTRAP='USB_A20'
   # Delays below are in seconds.
   _BOOTSTRAP_DELAY=0.1
-  _FIRST_FRAME_DELAY=0.1
-  _INTER_FRAME_DELAY=0.01
+  _FIRST_FRAME_DELAY=0.2
+  _INTER_FRAME_DELAY=0.02
```
alphan, line 130: just curious: did you adjust these just to be on the safe side (no need to revert) or were these tight already and stopped working after the frame size update?

alphan (approved):
LGTM, thanks @vogelpi.

cw/cw305/util/spiflash.py
vogelpi: These delays were indeed too small for the increased frame size. But I didn't experiment to reduce them to the minimum.

vogelpi:
Thanks for the quick review @alphan , I am merging this now.

vogelpi:
Ah wait, no. The bigger frame size would probably break the current binaries. Let's wait until we have new binaries first.

alphan:
> Ah wait, no. The bigger frame size would probably break the current binaries. Let's wait until we have new binaries first.

Good point, it is nice to have commits that work and should be easy to achieve in this case.

vogelpi:
I am closing this PR as the relevant commit has been merged into #43. It's better to do one atomic change (re-align bitstream, binary and spiflash config).

PR #42, opened 2021-02-23
Realign top-level name

vogelpi:
This PR updates the getting started guide and capture config, and renames the bitstream name to realign the repo with OpenTitan. We have switched to a separate top-level configuration for SCA already a while ago.


alphan (approved):

vogelpi:
Thanks @alphan for your review! Rebased on master, merging now.

PR #45, opened 2021-02-26
Add ChipWhisperer-Lite support for batch captures

alphan:
This change adds ChipWhisperer-Lite support for batch captures, i.e. captures that use the 'b' (batch encrypt) command.

Capturing 10000 traces at ~870 traces/s:
```
$ ./simple_capture_traces_batch.py -s cw_lite
Connecting and loading FPGA... SKIPPED! (see: /repo/cw/cw305/util/device.py:initialize_fpga)
Initializing PLL1
Programming OpenTitan with "objs/aes_serial_fpga_nexysvideo.bin"...
Transferring frame 0x00000000 @ 0x00000000.
Transferring frame 0x00000001 @ 0x000007D8.
Transferring frame 0x00000002 @ 0x00000FB0.
Transferring frame 0x00000003 @ 0x00001788.
Transferring frame 0x00000004 @ 0x00001F60.
Transferring frame 0x00000005 @ 0x00002738.
Transferring frame 0x80000006 @ 0x00002F10.
Serial baud rate = 38400
Serial baud rate = 115200
Scope setup with sampling rate 100003074.0 S/s
Target simpleserial version: z01 (attempts: 2).
WARNING:root:segment mode requires (samples-1) divisible by 3, rounding up to 1...
WARNING:root:segment mode requires (samples-1) divisible by 3, rounding up to 742...
Connected to ChipWhisperer-Lite (num_samples: 740, num_samples_actual: 741, num_segments_actual: 34)
Using key: '2b7e151628aed2a6abf7158809cf4f3c'.
Capturing: 100%|████████████████████| 10000/10000 [00:11<00:00, 877.87 traces/s]
```
image

While I still need to do some more cleanup and refactoring (see TODOs in the code), I wanted to share this as soon as possible. Some things worth mentioning:
* This change requires a newer bitstream for cw-lite (already on the develop branch of chipwhisperer), which you can get by running `pip install -r REQUIREMENTS_FILE`.
* `CwLiteSegmented._parse_waveform` uses `as_strided` to reshape the data received from cw-lite with minimal overhead.
* It looks like a lean interface for scopes is coming together (arm, capture, num_segments, num_samples, etc.). I hope to add this in a separate PR soon.
* The capture rate is slower than my initial trials (~1300 traces/s for 180 samples/trace) because those were done using the older bitstream. The number of samples per trace is 740 now.

@vogelpi and @vrozic, I didn't have the chance to try this thoroughly. It would be great if you could also give this a try and let me know if you run into any issues.
@moidx, fyi.

### DIFF
--- a/cw/cw305/cw_lite_segmented.py
+++ b/cw/cw305/cw_lite_segmented.py
@@ -1,0 +1,181 @@
+# Copyright lowRISC contributors.
+# Licensed under the Apache License, Version 2.0, see LICENSE for details.
+# SPDX-License-Identifier: Apache-2.0
+
+"""Support for capturing traces using ChipWhisperer-Lite in segmented mode."""
+
+import logging
+import re
+import time
+
+import chipwhisperer as cw
+import numpy as np
+from numpy.lib.stride_tricks import as_strided
+
+
+class CwLiteSegmented:
+    """Class for capturing traces using a ChipWhisperer-Lite.
+
+    This class uses segmented traces mode to improve capture performance.
+
+    When in segmented mode, ChipWhisperer-Lite captures multiple segments each starting
+    at a trigger event. This is much more efficient than sending a separate command for
+    each segment.
+
+    Due to current limitations of the firmware, ChipWhisperer-Lite needs to fill the
+    entire sample buffer when running in batch mode. Therefore, the number of segments
+    that need to be captured may be fewer or more than the desired number of segments
+    depending on the number of samples per segment. Additionally, the last segment may
+    be captured only partially. This class handles all the associated processing,
+    including discarding partially captured segments, internally.
+
+    However, the target must generate the correct number of triggers so that the entire
+    sample buffer is filled. This number is made available via the read-only
+    ``num_segments_actual`` attribute so that the target can be configured accordingly.
+
+    Typical usage:
+    >>> cw_lite = CwLiteSegmented()
+    >>> while foo:
+    >>>     ...
+    >>>     cw_lite.num_segments = desired_num_segments
+    >>>     # Note: cw_lite.num_segments_actual gives the actual number of segments that
+    >>>     # will be captured.
+    >>>     cw_lite.arm()
+    >>>     # Configure the target using cw_lite.num_segments_actual.
+    >>>     target.configure(cw_lite.num_segments_actual)
+    >>>     # This returns cw_lite.num_segments number of waves each with
+    >>>     # cw_lite.num_samples samples.
+    >>>     waves = cw_lite.capture_and_transfer_waves()
+    >>>     ...
+
+    Attributes:
+        num_samples_min: Minimum number of samples per segment. Read-only.
+        num_samples_max: Maximum number of samples per segment. Read-only.
+        num_samples: Number of samples per segment, must be in [``num_samples_min``,
+            ``num_samples_max``].
+        num_segments_min: Minimum number of segments per capture. Read-only.
+        num_segments_max: Maximum number of segments per capture, depends on the number
+            of samples per segment. Read-only.
+        num_segments: Number of segments per capture, must be in [``num_segments_min``,
+            ``num_segments_max``].  This number determines the number of segments
+            returned by ``capture_and_transfer_segments``.
+        num_segments_actual: Actual number of segments that will be captured. This
+            number depends on the number of samples per segment. The target must create
+            this many trigger events. Read-only.
+    """
+
+    def __init__(self, num_samples=740):
+        """Inits a CwLiteSegmented.
+
+        Args:
+            num_samples: Number of samples per segment, must be in [``num_samples_min``,
+                ``num_samples_max``].
+        """
+        self._scope = cw.scope()
+        self._configure_scope()
+        self.num_segments = 1
+        self.num_samples = num_samples
+        self._print_device_info()
+
+    @property
+    def num_segments_min(self):
+        return 1
+
+    @property
+    def num_segments_max(self):
+        return (self._scope.adc.oa.hwMaxSamples // self._scope.adc.samples) - 1
+
+    @property
+    def num_segments_actual(self):
+        # Must round-up to fill the entire buffer.
+        return round(self._scope.adc.oa.hwMaxSamples / self._scope.adc.samples) + 1
+
+    @property
+    def num_segments(self):
+        return self._num_segments
+
+    @num_segments.setter
+    def num_segments(self, num_segments):
+        if not self.num_segments_min <= num_segments <= self.num_segments_max:
+            raise RuntimeError(
+                f"num_segments must be in [{self.num_segments_min}, {self.num_segments_max}]."
+            )
+        self._num_segments = num_segments
+
+    @property
+    def num_samples_min(self):
+        return 1
+
+    @property
+    def num_samples_max(self):
+        # TODO: is this correct?
+        return self._scope.adc.oa.hwMaxSamples
+
+    @property
+    def num_samples(self):
+        return self._num_samples
+
+    @num_samples.setter
+    def num_samples(self, num_samples):
+        if not self.num_samples_min <= num_samples <= self.num_samples_max:
+            raise RuntimeError(
+                f"num_samples must be in [{self.num_samples_min}, {self.num_samples_max}]."
+            )
+        self._num_samples = num_samples
+        # This should ideally be handled by the chipwhisperer library but setting the
+        # number of samples smaller than 241 results in "received fewer points than
+        # expected" error. This number is further rounded up by chipwhisperer so that
+        # (num_samples-1) is divisible by 3. We get the actual number of samples from
+        # adc.
+        self._scope.adc.samples = max(241, num_samples)
+        # Note: CW-Lite actually returns one less than adc.samples.
+        self._num_samples_actual = self._scope.adc.samples - 1
+        if self.num_segments > self.num_segments_max:
+            print(f"Warning: Adjusting number of segments to {self.num_segments_max}.")
+            self.num_segments = self.num_segments_max
+
+    def _configure_scope(self):
+        self._scope.gain.db = 23
+        self._scope.adc.offset = 0
+        self._scope.adc.basic_mode = "rising_edge"
+        self._scope.adc.fifo_fill_mode = "segment"
+        self._scope.clock.clkgen_freq = 100000000
+        # We sample using the target clock (100 MHz).
+        self._scope.clock.adc_src = "extclk_dir"
+        self._scope.trigger.triggers = "tio4"
+        self._scope.io.tio1 = "serial_tx"
+        self._scope.io.tio2 = "serial_rx"
+        self._scope.io.hs2 = "disabled"
+        self._scope.clock.reset_adc()
+        time.sleep(0.5)
+        assert self._scope.clock.adc_locked, "ADC failed to lock"
+
+    def _print_device_info(self):
+        print(
+            (
+                "Connected to ChipWhisperer-Lite ("
+                f"num_samples: {self.num_samples}, "
+                f"num_samples_actual: {self._num_samples_actual}, "
+                f"num_segments_actual: {self.num_segments_actual})"
+            )
+        )
+
+    def arm(self):
+        """Arms ChipWhisperer-Lite."""
+        self._scope.arm()
+
+    def _parse_waveform(self, data):
+        shape = (self.num_segments, self.num_samples)
+        strides = (data.itemsize * self._num_samples_actual, data.itemsize)
+        return as_strided(data, shape, strides, writeable=False)
+
+    def capture_and_transfer_waves(self):
+        """Waits until the acquisition is complete and transfers waves.
+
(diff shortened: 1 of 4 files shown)


vogelpi:
This looks awesome @alphan ! I tested it on my machine and got around 720 traces/s. I don't worry about the somewhat lower performance compared to your measurement. It is probably due to my machine (AMD Ryzen). In non batch mode, I also get a bit less compared to when using my old Intel laptop.

One thing to note if you try this out: you should unplug/replug the CW-Lite after updating the firmware. Otherwise it runs with the old firmware and you don't get meaningful output.

Sidenote: I've talked to @vrozic today and it seems his CW-Lite is bricked. So he might not be able to try this out at the moment.

I will have a look at the code tomorrow and provide more feedback.

vogelpi:
Another thing that came to my mind when looking at the captured traces: should we do a longer capture with an AES configuration that we know can be broken to evaluate if this change in setup/alignment/data reshaping doesn't negatively affect signal quality?

colinoflynn:
@vogelpi - on speed - There may be some variation due to USB host controller (I forget the difference between the two main ones & how it interacts with the SW) too.

@vrozic - The CW-Lites should be "unbrickable" as they have a ROM bootloader - you can short the ERASE pins on the board & it should come up as a serial port, in which can you can run the instructions at [Firmware Update](https://chipwhisperer.readthedocs.io/en/latest/api.html#firmware-update). Otherwise let us know & we can do an advance replacement to get this back up asap.

One open point - we can setup some HW for people to run stuff on as well.

cw/cw305/simple_capture_traces_batch.py
```diff
@@ -47,17 +71,18 @@ def run_batch_capture(capture_cfg, ot, ktp):
...
             # Start batch encryption.
-            ot.target.simpleserial_write("b", num_waves.to_bytes(4, "little"))
+            ot.target.simpleserial_write(
+                "b", scope.num_segments_actual.to_bytes(4, "little")
+            )
             # Transfer traces
-            waves = waverunner.wait_for_acquisition_and_transfer_waves()
-            assert waves.shape[0] == num_waves
+            waves = scope.capture_and_transfer_waves()
+            assert waves.shape[0] == scope.num_segments
```
vogelpi, line 82: Not 100% sure but I think you should use `num_segments_actual` here.

cw/cw305/simple_capture_traces_batch.py
```diff
@@ -80,14 +105,31 @@ def run_batch_capture(capture_cfg, ot, ktp):
                     cw.common.traces.Trace(wave, plaintext, bytearray(ciphertext), key)
                 )
             # Update the loop variable and the progress bar.
-            rem_num_traces -= num_waves
-            pbar.update(num_waves)
+            rem_num_traces -= scope.num_segments
+            pbar.update(scope.num_segments)
```
vogelpi, line 109: Also here I think you should use `num_segments_actual`.

cw/cw305/cw_lite_segmented.py
```diff
@@ -0,0 +1,175 @@
...
+        self.num_samples = num_samples
+        self._print_device_info()
+
+    @property
+    def num_segments_min(self):
+        return 1
+
+    @property
+    def num_segments_max(self):
+        return (self._scope.adc.oa.hwMaxSamples // self._scope.adc.samples) - 1
```
vogelpi, line 80: Is the -1 really needed?

cw/cw305/cw_lite_segmented.py
```diff
@@ -0,0 +1,175 @@
...
+
+    @property
+    def num_segments_actual(self):
+        # Must round-up to fill the entire buffer.
```
vogelpi, line 84: This is not clear to me but I think I now start to understand.

It seems to me that we actually want to capture `num_segments_max` segments per batch which with the current settings would be floor(24573/742) - 1 = 32. This is the maximum number of complete segments that CW-Lite can hold in its sample buffer.

If I understand it correctly, we must - due to some limitations in the firmware - always fill the complete sample buffer. I.e., after the 32 segments we collect more samples until the buffer is full. Since this last segment is not captured completely, we will discard it.

Is that correct? if yes, it would be great if you could clarify a bit the comments.

cw/cw305/cw_lite_segmented.py
```diff
@@ -0,0 +1,175 @@
...
+        return round(self._scope.adc.oa.hwMaxSamples / self._scope.adc.samples) + 1
```
vogelpi, line 85: Can't you just do `math.ceil(self._scope.adc.oa.hwMaxSamples / self._scope.adc.samples)`?

cw/cw305/cw_lite_segmented.py
vogelpi: Wouldn't `math.floor(self._scope.adc.oa.hwMaxSamples / self._scope.adc.samples)` be sufficient here?

cw/cw305/simple_capture_traces_batch.py
vogelpi: After having reviewed more code, I think `num_segments` is correct here.

cw/cw305/simple_capture_traces_batch.py
vogelpi: Also here, I think `num_segments` is actually correct.

vogelpi (commented):
Thanks @alphan for working on this! That's cool stuff. I've left some comments.

vogelpi:
Thanks @colinoflynn for your input, we'll be in contact on Slack.

vrozic:
Thanks @colinoflynn - this solved the problem.
After short-circuiting the erase pin I was able to update the firmware using BOSSA

cw/cw305/simple_capture_traces_batch.py
```diff
@@ -80,14 +105,31 @@ def run_batch_capture(capture_cfg, ot, ktp):
...
+        Supports both ChipWhisperer-Lite and WaveRunner 9104."""
+    )
+    parser.add_argument(
+        "-s",
+        "--scope",
+        choices=["cw_lite", "waverunner"],
+        required=True,
+        help="scope to use",
+    )
+    return parser.parse_args()
```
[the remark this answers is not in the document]
alphan: Yes, this currently uses `capture.yaml` like other scripts.

cw/cw305/simple_capture_traces_batch.py
```diff
@@ -17,15 +24,35 @@
...
+def create_waverunner(ot, capture_cfg):
+    return WaveRunner(capture_cfg["waverunner_ip"])
+
+
+def create_cw_lite_segmented(ot, capture_cfg):
+    # TODO: Remove this disconnect after removing cw-lite init from device.py.
+    ot.scope.dis()
+    # Default samples per trace - We oversample by 10x and AES with DOM is doing
+    # ~56/72 cycles per encryption (AES-128/256).
+    return CwLiteSegmented(num_samples=740)
```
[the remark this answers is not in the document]
alphan: I agree.

cw/cw305/cw_lite_segmented.py
alphan: Based on my trials we need `-1` because without it the last trace seems to be missing some samples (when number of samples is 300):

image

@colinoflynn may be able to provide more insight but this could be an accuracy issue in `hwMaxSamples`.

The effect of this on the capture performance should be negligible anyway so I think we should keep it until things stabilize.

cw/cw305/cw_lite_segmented.py
alphan: It works for shorter traces but when the number of samples > 7000, we get a timeout error. Similar to above, better safe than sorry :)

```
...
Connected to ChipWhisperer-Lite (num_samples: 7000, num_samples_actual: 7002, num_segments_actual: 3)
Using key: '2b7e151628aed2a6abf7158809cf4f3c'.
Capturing:   0%|                                  | 0/5000 [00:00<?, ? traces/s]
WARNING:root:Timeout in OpenADC capture(), no trigger seen! Trigger forced, data is invalid. Status: 08
...
```

cw/cw305/cw_lite_segmented.py
alphan: Yes, your understanding is correct but let me rephrase for the sake of clarity: CW-Lite always captures `num_segments_actual` segments, this depends on the number of samples per trace and the target must create this many trigger events. `num_segments` can be anything in `[num_segments_min, num_segments_max]` and is basically a convenience so that we can encapsulate post-processing in this class.

Can you take another look at the docstring of the class (I thought that along with this comment was clear enough)?

cw/cw305/cw_lite_segmented.py
```diff
@@ -0,0 +1,175 @@
...
+        # (num_samples-1) is divisible by 3. We get the actual number of samples from
+        # adc.
+        self._scope.adc.samples = max(241, num_samples)
+        # Note: CW-Lite actually returns one less than adc.samples.
+        self._num_samples_actual = self._scope.adc.samples - 1
+        if self.num_segments > self.num_segments_max:
+            print(f"Warning: Adjusting number of segments to {self.num_segments_max}.")
+            self.num_segments = self.num_segments_max
+
+    def _configure_scope(self):
```
moidx, line 131: nit: Maybe as a TODO, but it would be good to extract this information, maybe the class should take in a scope instance? Construction wise I think we are doing this in other places.

moidx (approved):
LGTM. Thanks @alphan!

cw/cw305/cw_lite_segmented.py
vogelpi: Ah, I see. Thanks for the explanation.

cw/cw305/cw_lite_segmented.py
```diff
@@ -0,0 +1,175 @@
...
+
+    When in segmented mode, ChipWhisperer-Lite captures multiple waves each starting at
+    a trigger event. This is much more efficient than sending a separate command for
+    each wave.
+
+    Due to current limitations of the hardware, number of waves (segments) per capture
+    is fixed and there is also a lower limit on the number of samples per wave. This
+    class handles this internally but also makes this information available via the
+    read-only ``num_segments_actual`` attribute so that the target can be configured to
+    generate the correct number of triggers.
```
vogelpi, line 29: I would clarify this section e.g. as follows:

"Due to current limitations of the firmware, ChipWhisperer-Lite always needs to fill the entire sample buffer when running in batch mode. Depending on the number of samples per segment, fewer or more segments need to be captured to fill the sample buffer, and the last segment may be captured only partially. Internally, this class handles all the associated processing including discarding partially captured segments.

However, it must be ensured that the target indeed generates the correct number of capture triggers to fill the sample buffer. This number is made available via the read-only ``num_segments_actual`` attribute so that the target can be configured accordingly."

cw/cw305/cw_lite_segmented.py
vogelpi: Thanks for confirming. I did read the docstring but it wasn't clear to me. I have now made a suggestion how to clarify the docstring.

vogelpi (approved):
Thanks for your feedback @alphan . I think the docstring of the `CwLiteSegmented` can be improved to clarify what's happening. I've made a suggestion. Apart from that, this PR looks good to me!

cw/cw305/cw_lite_segmented.py
```diff
@@ -0,0 +1,175 @@
...
+    def __init__(self, num_samples=740):
+        """Inits a CwLiteSegmented.
+
+        Args:
+            ip_addr: IP address of the oscilloscope.
+        """
+        self._scope = cw.scope()
+        self._configure_scope()
+        self.num_segments = 1
+        self.num_samples = num_samples
```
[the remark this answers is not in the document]
alphan: Done, thanks for catching this.

cw/cw305/cw_lite_segmented.py
alphan: I agree. There are several ways to address this. I already have a related TODO in `simple_capture_traces_batch.py` to remove scope init from `device.py`. The main differences between the batch and single capture modes are `basic_mode` and `fifo_fill_mode`, which we can handle by adding a base class for both modes of cw-lite.

cw/cw305/cw_lite_segmented.py
alphan: Done, thanks.

alphan:
> Another thing that came to my mind when looking at the captured traces: should we do a longer capture with an AES configuration that we know can be broken to evaluate if this change in setup/alignment/data reshaping doesn't negatively affect signal quality?

IIRC, I was able to recover the key using an older bitstream. I can give this another try when I have the chance.

vogelpi:
> > Another thing that came to my mind when looking at the captured traces: should we do a longer capture with an AES configuration that we know can be broken to evaluate if this change in setup/alignment/data reshaping doesn't negatively affect signal quality?
> 
> IIRC, I was able to recover the key using an older bitstream. I can give this another try when I have the chance.

Excellent, thanks @alphan!

PR #46, opened 2021-06-16
Maintain high capture rate for long batch captures

vogelpi:
This PR contains two commits in order to increase the capture rate for long batch captures.
1. Frequent storage segment array resizing is avoided by setting the hint value to the known storage segment size (a project parameter).
2. During capture, `disable` all except for the latest two storage segments to keep the run time of the `append()` function approximately constant. By default, it increases linearly with the number of traces captured.

With these changes it is possible to maintain high capture rates up to millions of traces. Further changes are probably needed to avoid memory limitations. Currently everything is kept in memory during a capture and traces are written to disk only at the very end. I've opened an issue newaetech/chipwhisperer#344 in the hope of getting some further advice.

### DIFF
--- a/cw/cw305/simple_capture_traces_batch.py
+++ b/cw/cw305/simple_capture_traces_batch.py
@@ -97,6 +97,13 @@ def run_batch_capture(capture_cfg, ot, ktp, scope):
                 f"actual: {actual_last_ciphertext}\n"
                 f"expected: {expected_last_ciphertext}"
             )
+            # Make sure to allocate sufficient memory for the storage segment array during the
+            # first resize operation. By default, the ChipWhisperer API starts every new segment
+            # with 1 trace and then increases it on demand by 25 traces at a time. This results in
+            # frequent array resizing and decreasing capture rate.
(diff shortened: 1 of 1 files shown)


cw/cw305/simple_capture_traces_batch.py
```diff
@@ -97,17 +98,34 @@ def run_batch_capture(capture_cfg, ot, ktp, scope):
                 f"actual: {actual_last_ciphertext}\n"
                 f"expected: {expected_last_ciphertext}"
             )
+            # Make sure to allocate sufficient memory for the storage segment array during the
```
moidx, line 101: Would it make sense to add a reference to the CW issue here? newaetech/chipwhisperer#344

moidx (approved):

alphan (approved):

cw/cw305/simple_capture_traces_batch.py
vogelpi: Done.

vogelpi:
Thanks for the quick review guys!

PR #51, opened 2021-08-17
Add SHA3/KMAC capture support using Typer

alphan:
This PR adds initial support for capturing sha3/kmac traces and adds `capture aes,` `capture sha3`, `init`, and `plot` commands to the CLI using the `typer` library. This change also refactors device initialization a bit by moving fpga and programmer selection to `device.py` and adds CW310 binaries from lowRISC/opentitan@b1188ef.

This PR includes the following commits:
* Move fpga and programmer selection to device.py
* Add CW310 bitstream and binaries from lowRISC/opentitan@b1188ef
* Use separate config files for aes and sha3 (please don't forget to use `--cfg-file capture_sha3.yaml` for sha3)
* Move output_len to OpenTitan() init
* Add SHA3/KMAC capture support using Typer

I verified that aes capture works on both boards and sha3 capture works on CW310 but I would appreciate it if you could try this change on your setup and let me know if you run into any issues.

Command line help messages:
```
$ ./simple_capture_traces.py --help
Usage: simple_capture_traces.py [OPTIONS] COMMAND [ARGS]...

  Capture traces for side-channel analysis.

Options:
  --cfg-file TEXT
  --help           Show this message and exit.

Commands:
  capture  Capture traces for SCA
  init     Initalize target for SCA.
  plot     Plots previously captured traces.
```

```
$ ./simple_capture_traces.py capture --help
Usage: simple_capture_traces.py capture [OPTIONS] COMMAND [ARGS]...

  Capture traces for SCA

Options:
  --help  Show this message and exit.

Commands:
  aes   Capture AES traces from a target that runs the `aes_serial` program.
  sha3  Capture KMAC traces from a target that runs the `sha3_serial`...
```

```
$ ./simple_capture_traces.py capture aes --help
Usage: simple_capture_traces.py capture aes [OPTIONS]

  Capture AES traces from a target that runs the `aes_serial` program.

Options:
  --num-traces INTEGER   Number of traces to capture.
  --plot-traces INTEGER  Number of traces to plot.
  --help                 Show this message and exit.
```

### DIFF
--- a/cw/cw305/simple_capture_traces.py
+++ b/cw/cw305/simple_capture_traces.py
@@ -23,7 +23,8 @@ def initialize_capture(device_cfg, capture_cfg):
                           device_cfg['pll_frequency'],
                           device_cfg['baudrate'],
                           capture_cfg['scope_gain'],
-                          capture_cfg['num_samples'])
+                          capture_cfg['num_samples'],
+                          capture_cfg['plain_text_len_bytes'])
     print(f'Scope setup with sampling rate {ot.scope.clock.adc_rate} S/s')
     # Ping target
     print('Reading from FPGA using simpleserial protocol.')
@@ -132,7 +133,6 @@ if __name__ == "__main__":
     ktp = cw.ktp.Basic()
     ktp.key_len = cfg_file['capture']['key_len_bytes']
     ktp.text_len = cfg_file['capture']['plain_text_len_bytes']
-    ot.target.output_len = cfg_file['capture']['plain_text_len_bytes']
 
     run_capture(cfg_file['capture'], ot, ktp)
 
--- a/cw/cw305/simple_capture_traces.py
+++ b/cw/cw305/simple_capture_traces.py
@@ -9,36 +9,21 @@ import numpy as np
 import time
 from tqdm import tqdm
 import yaml
-import re
 
 import chipwhisperer as cw
 
 from util import device
 from util import plot
-from vendor.lowrisc_opentitan import cw_spiflash
 
 
 def initialize_capture(device_cfg, capture_cfg):
     """Initialize capture."""
-    # Extract target board type from bitstream name.
-    m = re.search('cw305|cw310', device_cfg['fpga_bitstream'])
-    if m:
-        if m.group() == 'cw305':
-            board = 'CW305'
-        else:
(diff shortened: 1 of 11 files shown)


cw/cw305/simple_capture_traces.py
```diff
@@ -40,101 +55,153 @@ def initialize_capture(device_cfg, capture_cfg):
...
+    # Key and plaintext generator
+    ctx.obj.ktp = cw.ktp.Basic()
+    ctx.obj.ktp.key_len = cfg.capture.key_len_bytes
+    ctx.obj.ktp.text_len = cfg.capture.plain_text_len_bytes
+
+    ctx.obj.ot = initialize_capture(cfg.device, cfg.capture)
 
-        ot.target.flush()
 
+def capture_loop(traces, capture_cfg):
```
vogelpi, line 105: If I understand this correctly, `traces` is actually a function including arguments. Inside the loop, you then just call that over and over again. I like this as it allows for sharing the logic around for mutliple capture functions.

But what I don't understand:
- What is `next()` doing on a function? Is it just running it once more, or is there some more magic going on? 
- Inside the `traces` function, you are updating `ktp`. Does that mean every trace is collected with a different key?

Suggestions:
- I would give `traces` a different name to indicate that it's actually a function capturing a trace.
- Maybe add another, optional function to initialize the reference cipher such that it doesn't have to be done inside the capture function upon every call.

vogelpi (commented):
Hey @alphan,
this looks great! I like the idea of the much more versatile CLI and that infrastructure can be shared between KMAC and AES. Also, moving the bitstream flashing and binary loading into the device setup is very reasonable.

Since this is still a draft, what is your envisioned way forward? Are there more things you plan to add? Or do you want to extend this to the segmented capture as well before merging? Just curious...

cw/cw305/simple_capture_traces.py
vogelpi: These explanations are very helpful, thanks @alphan! (also for renaming `traces` to `trace_gen`).

vogelpi (approved):
It took me a while, but I've finally also managed to try this out on the FPGA. LGTM, thanks again for the work you've put into this @alphan !

Issue #30, opened 2021-01-18
Report the number of correctly guessed differences in ceca.py.


alphan:
See this [comment](https://github.com/lowRISC/ot-sca/pull/28#discussion_r545755844) in #28.

alphan:
For all 120 differences.

(We can also report the number of correct key bytes as well but based on my observation it doesn't provide a lot of information.)

@vogelpi, fyi.

vogelpi:
This got resolved with #53 .

PR #53, opened 2021-08-20
[ceca] Report number of correctly guessed key byte differences

vogelpi:
This number is a actually a far more relevant indicator for the amount of leakage in the traces than successful key recovery (pass/fail). Therefore, this PR adds the functionality to compute this number and report it on the command line.

This solves #30 .

### DIFF
--- a/cw/cw305/ceca.py
+++ b/cw/cw305/ceca.py
@@ -82,7 +82,6 @@ class TraceWorker:
         """
         self.project = cw.open_project(project_file)
         # TODO: Consider more efficient formats.
-        trace_cnt = trace_slice.stop - trace_slice.start
         self.num_samples = attack_window.stop - attack_window.start
         if attack_direction == AttackDirection.INPUT:
             self.texts = np.vstack(self.project.textins[trace_slice])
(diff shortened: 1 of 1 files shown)


cw/cw305/ceca.py
```diff
@@ -408,6 +407,34 @@ def recover_key(diffs, attack_direction, plaintext, ciphertext):
...
+    """Compares pairwise_diffs_scores with diffs between bytes in correct_key.
+
+    This function takes the differences between key bytes computed in
+    pairwise_diffs_scores and compares them with the actual differences in the
+    correct key.
+
+    Args:
+        pairwise_diffs_scores: A 16x16x2 matrix of pairwise differences between
+            key bytes and their confidence scores.
+        attack_direction: Attack direction, 'input' or 'output'.
```
alphan, line 420: nit: "'input' or 'output'" reads like `attack_direction` is a string.

alphan (approved):
Thanks @vogelpi!

cw/cw305/ceca.py
vogelpi: Hey Alphan, thanks for your review!

There are too many quotes here, sorry. How exactly would you like me to change this?

alphan: Ah, sorry, I was thinking maybe we should remove that part:

```suggestion
        attack_direction: Attack direction.
```

vogelpi: Oh, I see. I can do that. Thanks for the clarification.

PR #54, opened 2021-09-07
KMAC support for Python scripts

m-temp:
This is WIP to add KMAC (and cshake) support to the SCA-repo.
Unfortunately, there is no (native) support for kmac and cshake (mainly because of the customization string in SP800-185) in python. There is a [feature request](https://bugs.python.org/issue39539) for hashlib, but to the best of my knowledge there is no SP800-185 python module available. Therefore I used the standard (other optimizations would be possible) C-implementation from the official [eXtended Keccak Code Package](https://github.com/XKCP/XKCP) which uses the [CC0](https://creativecommons.org/publicdomain/zero/1.0/) and a BSD 3-clause license. Is this OK? 

The provided demo works for the testvektors and are compatible to cw's ktp-generator.

As I'm waiting for a cw310 board I cannot integrate the code into the ot-sca-framework atm.
However this pull request is intended to further discuss the implementation process. Help on how to integrate the cffi, ffi/lib stuff into a package is welcome. Apologies, if I missed things in the integration flow. Please bear with me, I'm new ;)


cw/cw305/kmac_cw_demo.py
```diff
@@ -0,0 +1,61 @@
+import argparse
```
vogelpi, line 1: Please add the OpenTitan license header to all files that you wrote yourself. You can find the license header here: https://docs.opentitan.org/doc/project/detailed_contribution_guide/#writing-code

cw/cw305/pyXKCP/KeccakP-1600-SnP.h
```diff
@@ -0,0 +1,40 @@
+/*
```
vogelpi, line 1: If I understand it correctly, most of the code under `pyXKCP` is actually from the XKCP project. It is okay to include code from other projects with compatible licenses but in OpenTitan we use the vendor tool to include such code: https://docs.opentitan.org/doc/project/detailed_contribution_guide/#vendored-code

As an example, we have used this in ot-sca to vendor in the `cw_spiflash.py` under `cw/cw305/vendor/lowrisc_opentitan` from the main OpenTitan repository. The advantage of using the vendor tool is that at any point in time, we can update the copy in the tree from upstream. the lockfile (lowrisc_opentitan.lock.hjson) is used to remember at which point in time we did the last update. Any changes, we need apply to the copy in tree are handled through patches. Please have a look at this. Further examples can be found in OpenTitan, see opentitan/hw/vendor. If you need more information, just let me know. I am happy to help.

vogelpi: What I forgot to mention, you can also just specify a limited set of sources that the tool should include. This is useful in case you just need one file from a big project, e.g., `cw_spiflash.py`.

vogelpi (changes requested):
Thanks @m-temp for your first PR to ot-sca! ;-)

At the moment, it's not fully clear to me what code is from XKCP and what code you contributed yourself, i.e., where I should focus with my review. I suggest that you try to use the OpenTitan vendor tool for including the XKCP sources and then things become more obvious for me and I can do a second pass.

Please just let me know if things are not clear or you need some help.

cw/cw305/pyXKCP/KeccakP-1600-SnP.h
m-temp: Thank you for pointing this out. I will rework the code to be compliant with OT's vendor policy.

m-temp (commented):
I reworked the code as suggested. However, I need to place this single config file in the vendor directory, or patch all files to use a different location.
 What's the best way to do it?
I tried to use the `patch_dir` config option, however it seems like the patch_dir can only be used for "real" patches, but not for creating files. `CRITICAL: In hjson at cw/cw305/vendor/xkcp_xkcp.vendor.hjson, Has patch_dir, but no mapping item uses it.`

cw/cw305/kmac_cw_demo.py
m-temp: done

cw/cw305/pyXKCP/pyXKCP_extention_build.py
```diff
@@ -0,0 +1,95 @@
...
+#int TupleHash128_Update(TupleHash_Instance *TupleHashInstance, const TupleElement *tuple, size_t numberOfElements);
+#int TupleHash128_Final(TupleHash_Instance *TupleHashInstance, BitSequence * output);
+#int TupleHash128_Squeeze(TupleHash_Instance *TupleHashInstance, BitSequence *output, BitLength outputBitLen);
+#int TupleHash256( const TupleElement *tuple, size_t numberOfElements, BitSequence *output, BitLength outputBitLen, const BitSequence *customization, BitLength customBitLen);
+#int TupleHash256_Initialize(TupleHash_Instance *TupleHashInstance, BitLength outputBitLen, const BitSequence *customization, BitLength customBitLen);
+#int TupleHash256_Update(TupleHash_Instance *TupleHashInstance, const TupleElement *tuple, size_t numberOfElements);
+#int TupleHash256_Final(TupleHash_Instance *TupleHashInstance, BitSequence * output);
+#int TupleHash256_Squeeze(TupleHash_Instance *TupleHashInstance, BitSequence *output, BitLength outputBitLen);
+
+ffibuilder.set_source("_XKCP",  # name of the output C extension #ToDo: better naming?
```
vogelpi, line 57: I don't really know about the naming "_XKCP" sounds reasonable to me. Maybe the [Python style guide](https://docs.opentitan.org/doc/rm/python_coding_style/) has some guidance on that? 

In fact, the [PEP 8 Style Guide](https://www.python.org/dev/peps/pep-0008/#package-and-module-names) says:
> Modules should have short, all-lowercase names. Underscores can be used in the module name if it improves readability. Python packages should also have short, all-lowercase names, although the use of underscores is discouraged.
>
> When an extension module written in C or C++ has an accompanying Python module that provides a higher level (e.g. more object oriented) interface, the C/C++ module has a leading underscore (e.g. _socket).

So, probably `_xkcp` should be used.

cw/cw305/pyXKCP/pyXKCP_extention_build.py
```diff
@@ -0,0 +1,95 @@
...
+    sources=['../vendor/xkcp_xkcp/SP800-185.c','../vendor/xkcp_xkcp/KeccakSponge.c','../vendor/xkcp_xkcp/KeccakP-1600-compact64.c'],
+    libraries=[])
+
+
+def create_config_file():
+    with open('../vendor/xkcp_xkcp/config.h', 'w') as the_file:
+#        the_file.write('// Copyright lowRISC contributors\n')
+#        the_file.write('// Licensed under the Apache License, Version 2.0, see LICENSE for details.\n')
+#        the_file.write('// SPDX-License-Identifier: Apache-2.0 \n\n')
+        the_file.write('// This file is autogenerated by ' +  __file__ + '\n// DO NOT CHANGE!\n\n')
```
vogelpi, line 77: Thanks for doing that :+1:

cw/cw305/pyXKCP/pyXKCP_extention_build.py
```diff
@@ -0,0 +1,95 @@
...
+""",
+    sources=['../vendor/xkcp_xkcp/SP800-185.c','../vendor/xkcp_xkcp/KeccakSponge.c','../vendor/xkcp_xkcp/KeccakP-1600-compact64.c'],
+    libraries=[])
+
+
+def create_config_file():
+    with open('../vendor/xkcp_xkcp/config.h', 'w') as the_file:
+#        the_file.write('// Copyright lowRISC contributors\n')
+#        the_file.write('// Licensed under the Apache License, Version 2.0, see LICENSE for details.\n')
+#        the_file.write('// SPDX-License-Identifier: Apache-2.0 \n\n')
```
vogelpi, line 76: It's a good idea to include this. We do that also in OpenTitan.

cw/cw305/pyXKCP/pyXKCP_extention_build.py
```diff
@@ -0,0 +1,95 @@
...
+        the_file.write('// This file is autogenerated by ' +  __file__ + '\n// DO NOT CHANGE!\n\n')
+        # Define which KeccakPermutation is used
+        the_file.write('#define XKCP_has_KeccakP1600\n')
+        the_file.close()
+
```
vogelpi, line 81: You have quite a few line breaks here :-)

What I would recommend is to run `flake8` to check the formatting and style. It will highlight such things. You can also have a look at [our Python style guide](https://docs.opentitan.org/doc/rm/python_coding_style/#lint-tool) on how to run a lint tool on Python. I think this is useful as it will lead to a more consisten style.

cw/cw305/pyXKCP/__init__.py
```diff
@@ -0,0 +1,5 @@
+# Copyright lowRISC contributors.
+# Licensed under the Apache License, Version 2.0, see LICENSE for details.
+# SPDX-License-Identifier: Apache-2.0
+
+from . import pyXKCP_extention_build
```
vogelpi, line 5: I am not a Python expert but I think the "right" way of doing this would be to have a completely empty `__init__.py` and put the  `from . import pyXKCP_extention_build` line into `cw/cw305/pyXKCP/pyXKCP.py.` Do you think that would work?

cw/cw305/pyXKCP/pyXKCP.py
```diff
@@ -0,0 +1,104 @@
...
+    ret = kmac128(key, keyLen, input, inputLen, outputLen, customization, customLen)
+    print("Output by value:      ", binascii.hexlify(ret))
+    if ret == expected_output:
+        print("Output by value matches. PASS!")
+    else:
+        print("Output by value does not match!")
+
+
+else:
+    from . import _XKCP
```
vogelpi, line 104: It's good to have such a test function but I would only keep it as a function definition in here and no main. Then a "user" including `pyXKCP`  can call that function in his/her own module.

cw/cw305/kmac_cw_demo.py
```diff
@@ -0,0 +1,65 @@
+# Copyright lowRISC contributors
```
vogelpi, line 1: I am not sure if this file would work at the moment (I mean the interfacing with the capture and target board etc.) and you probably also can't test this atm. To verify the XKCP interface, you have implemented a separate test function. So I am inclined to not merge this in the first place. What is your view?

vogelpi (commented):
Thanks for updating this PR and using the vendor tool @m-temp . I've left some comments around formatting and linting + some other stuff. But this is definitely going in the right direction.

One general comment: Since this is a PR, it is completely fine to fixup/squash commits and do force push. Ultimately, this PR should have one or two commits.

cw/cw305/kmac_cw_demo.py
m-temp: agree

cw/cw305/pyXKCP/pyXKCP.py
m-temp: ok. sounds resonable

cw/cw305/pyXKCP/__init__.py
m-temp: This should work, however, I am neither a Python expert.

cw/cw305/pyXKCP/pyXKCP_extention_build.py
```diff
@@ -0,0 +1,95 @@
+# Copyright lowRISC contributors.
```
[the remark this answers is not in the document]
m-temp: I agree

cw/cw305/pyXKCP/pyXKCP_extention_build.py
```diff
@@ -0,0 +1,95 @@
...
+#int TupleHash128_Initialize(TupleHash_Instance *TupleHashInstance, BitLength outputBitLen, const BitSequence *customization, BitLength customBitLen);
+#int TupleHash128_Update(TupleHash_Instance *TupleHashInstance, const TupleElement *tuple, size_t numberOfElements);
+#int TupleHash128_Final(TupleHash_Instance *TupleHashInstance, BitSequence * output);
+#int TupleHash128_Squeeze(TupleHash_Instance *TupleHashInstance, BitSequence *output, BitLength outputBitLen);
+#int TupleHash256( const TupleElement *tuple, size_t numberOfElements, BitSequence *output, BitLength outputBitLen, const BitSequence *customization, BitLength customBitLen);
+#int TupleHash256_Initialize(TupleHash_Instance *TupleHashInstance, BitLength outputBitLen, const BitSequence *customization, BitLength customBitLen);
+#int TupleHash256_Update(TupleHash_Instance *TupleHashInstance, const TupleElement *tuple, size_t numberOfElements);
+#int TupleHash256_Final(TupleHash_Instance *TupleHashInstance, BitSequence * output);
+#int TupleHash256_Squeeze(TupleHash_Instance *TupleHashInstance, BitSequence *output, BitLength outputBitLen);
```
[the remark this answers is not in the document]
m-temp: yeah, makes sense. When putting the lines there, I was thinking if it makes sense to support those functions, but I don't think that makes sense atm.

cw/cw305/pyXKCP/pyXKCP_extention_build.py
m-temp: done

cw/cw305/pyXKCP/pyXKCP_extention_build.py
m-temp: I'll use the lintpy.py from the opentitan repository as a pre-commit hook.

cw/cw305/pyXKCP/__init__.py
m-temp: I think the best way is either the __init__py in its current version, or require a manual build. What do you think?

vogelpi (approved):
Thanks @m-temp for addressing the feedback. I had to read up a bit on the `__init__.py` topic and I think calling the build function inside that is how this should be done. Sorry for raising this point in the first place. I also forgot to answer your question on the handling of the config file in the vendor dir. The solution you came up with looks okay to me, too.

Other than a couple of nits in the readme, this looks good to me, thanks for your work on this!

vogelpi:
This looks good, I've successfully tested this and integrated it into the capture (will file a PR shortly). I am merging this. Congrats @m-temp to your first PR :-)

Issue #50, opened 2021-08-17
Add support for CW-Husky


vogelpi:
Our capture scripts still use the CW-Lite scope. We should switch to CW-Husky which has higher capture performance.

colinoflynn:
I've also started adding some documentation here: https://github.com/newaetech/ot-sca/blob/cwhusky-doc/doc/getting_started_bergen.md

I'm using a separate branch for now to avoid polluting the active PR. If the documentation is done before that PR gets merged I'll move it over, but for now the commits will be a bit messy on the `cwhusky-doc` branch. However I wanted to get some notes out there ASAP, and if there is anything else useful please feel free to push into that.

m-temp:
> We have planned to restructure a bit the directory structure and with that should also modify cw305. No need to change that in your PR.

That's something we should definitively do.
Additionally, there seems to be some miss-configuration (on my side?) between git and git lfs:
When using the binaries from e64e76dfd09f8ac23ca8ba42465420e63e3c5381 I get `ValueError: Invalid file size: 41`
An additional `git lfs pull` provides new (?) different (?) binaries, that are marked as `modified`. However, I am not able to attack them. (0 to2 diffs recovered) Are they masked?

>I would expect the Husky to give better performance but the variation with the bitstream is big.

Building my own unmasked `aes_serial_fpga_cw310` from  lowRISC/opentitan@1b73f8090033839b73de6b5954702137d723948e and using a scope_gain of 35, I get the following:
```
./sweep_ceca.py -f projects/opentitan_simple_aes.cwp -s 3 -w 16 -d output -a 518 527
#traces
          key recovered
            #diffs recovered
    10000 0   0
   100000 0  14
   200000 0  52
   300000 0  80
   400000 0  93
   500000 1 104
   600000 1 110
   700000 1 112
   800000 1 111
   900000 1 114
  1000000 1 114
  1100000 1 116
  1200000 1 118
  1300000 1 119
  1400000 1 117

```
(Ran out of memory for more traces)

However, I have the impression that the measurements are not very reliable/reproducible. I have to explore this a little bit more.

vogelpi:
I am closing this issue as the PR for enabling CW-Husky has been merged a while ago already (#56 ). Btw, I've finally managed to try it out and it's great. I am really happy to have it, thanks @colinoflynn and @jpcrypt !

PR #56, opened 2021-10-26
Add CWHusky Support (#50)

colinoflynn:
This should add support for CW-Husky (#50). The default gain in the .yaml file should change for Husky, around 35 seemed to work for me (this could be changed in the file as well, not sure where makes sense). The CW-Lite gain is low but does work, so it's safer to keep cw-lite gain levels presumably in the code itself.

I can squash these if too many commits (especially somehow realising LFS added the full files so I had to revert them, which I'm not sure why that happened)


m-temp:
> I'll leave the squashing to the last step once everything else resolved.
> 
> On `python-requirements.txt` I guess that raises one particularly terrible point - CW-Lite + CW-Husky require different CW versions. I'll see if our next rebuild of the CW-Lite fixes segmenting again in which case they'd both work with latest.
> 
> But in the meantime I could unpin `python-requirements.txt` and also add a manual version check inside of the initialisation function that warns the user if they've got an incompatible CW version for now. This would at least reduce user frustration since I think errors from wrong version are very non-obvious (especially for someone just starting).

Hi,
first of all: thanks for the work on the husky support.
Unpinning the cw version currently leads to cw 5.6.1 which requires python3.7 (at least it says so in the setup.py), which is not the default in ubuntu 18.04 LTS and thus, requires a python update. This is for sure doable, but in general ubuntu 18.04 lts + python3.6 is recommended for opentitan. Is there a version that supports the husky and python3.6? Or can the `python_requires='~=3.7' `be softened?

colinoflynn:
Check with @alex-dewar - this is a follow-on from `numpy`, which started requiring 3.7 (https://numpy.org/neps/nep-0029-deprecation_policy.html). There is no critical numpy features we needed from latest, so you can just use whatever version of numpy comes with 3.6 and it should work.

m-temp (approved):
I could successfully reproduce the AES attack described at https://github.com/newaetech/ot-sca/blob/cwhusky-doc/doc/getting_started_bergen.md using the  the cw310 and husky.

One minor point:
- There is a API warning: `WARNING:ChipWhisperer Other:Could not check ChipWhisperer version, error __init__() got an unexpected keyword argument 'capture_output'`

PR #61, opened 2022-03-10
Documentation and dependency fixes

vogelpi:
This PR:
- Clarifies that Python 3.7 or later should be used.
- Fixes a typo in the doc.
- Fixes the ChipWhisperer version to something known good and installs the archive rather than using git checkout in editable mode (may lead to issues with the main OpenTitan install where we can't upgrade the ChipWhisperer toolchain anymore due to the Python 3.6.5 requirement of OT).

### DIFF
--- a/doc/getting_started.md
+++ b/doc/getting_started.md
@@ -63,7 +63,7 @@ machine or obtained using the provided [Dockerfile](https://github.com/lowRISC/o
 
 This repository has a couple of Python dependencies. You can run
 ```console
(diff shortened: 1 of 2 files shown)


vogelpi:
Hey @andreaskurth, as I don't have a board at the moment, could you start from this PR and tell me if everything works?

vogelpi:
@vrozic this might help you to get Husky working again.

vrozic (approved):
Indeed, this solves the problem with Husky.
Thanks for the update.

PR #62, opened 2022-03-18
[sca] Capture setup for general fixed-vs-random TVLA

vrozic:
Signed-off-by: vrozic <vrozic@lowrisc.org>

### DIFF
--- a/cw/cw305/simple_capture_traces.py
+++ b/cw/cw305/simple_capture_traces.py
@@ -2,7 +2,6 @@
 # Copyright lowRISC contributors.
 # Licensed under the Apache License, Version 2.0, see LICENSE for details.
 # SPDX-License-Identifier: Apache-2.0
-import argparse
 import binascii
 from Crypto.Cipher import AES
 import numpy as np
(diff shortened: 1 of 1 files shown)


andreaskurth (approved):
LGTM, thanks! I have been working with this for the measurements last week.

cw/cw305/simple_capture_traces.py
```diff
@@ -179,13 +180,60 @@ def capture_kmac(ot, ktp):
...
 
+def capture_kmac_key(ot, ktp):
+    """A generator for capturing KMAC traces.
+    In order to simplify the analysis, the first sample has to use fixed key.
+
+    Args:
+      ot: Initialized OpenTitan target.
+      ktp: Key and plaintext generator.
+    """
+    key, _ = ktp.next()
```
vogelpi, line 191: This key is never used as is the generator. So it could just be dropped. Alternatively, this generator could be more widely used instead of the fixed keys and texts + AES below. But then it wouldn't follow anymore the DTR spec.

vogelpi:
Ah, and could you please run `flake8 cw/cw305/simple_capture_traces.py --max-line-length 100` on this and fix the warnings?

PR #63, opened 2022-05-05
Batch capture for fast KMAC data aquisition

vrozic:
Enabled batch capture for KMAC data aquisition for TVLA general fixed-vs-random testing.

Signed-off-by: vrozic <vrozic@lowrisc.org>


vogelpi:
Thanks for the PR @vrozic , this is cool stuff! Question: would it be possible to integrate this into the existing batch capture script for AES? I had a quick look and it seems the two files are very similar. Merging them would allow us to reduce code duplication and maintenance overhead.

Also, can you please comment a little on what's still missing in the PR? I remember there were some issues that you needed to debug. Is everything solved now or do we still have some opens?

PR #65, opened 2022-05-09
Optimize SCA trace capture

vogelpi:
This PR contains three commits to optimize the trace capture using ChipWhisperer:
1. The sampling rate is increased to 200 MS/s which is the maximum supported by Husky. CW-Lite keeps collecting traces with 100 MS/s (max supported 105 MS/s) the number of samples to record is adjusted internally for CW-Lite.
2. Allow negative and positive sampling offsets to be defined through the config file. Is allows to better tune the capture window.
3. Tune the capture window for SHA3.

Also, this PR updates the ChipWhisperer version to get a more recent of the library.

### DIFF
--- a/cw/cw305/cw_segmented.py
+++ b/cw/cw305/cw_segmented.py
@@ -65,7 +65,7 @@ class CwSegmented:
             this many trigger events. Read-only.
     """
 
-    def __init__(self, num_samples=1200, scope_gain=23, scope=None):
+    def __init__(self, num_samples=1200, offset=0, scope_gain=23, scope=None):
         """Inits a CwSegmented.
 
         Args:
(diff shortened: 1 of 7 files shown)


vogelpi:
@johannheyszl This PR contains a first set of changes proposed in the updated Todo list. Next, I will try out to store the traces as integers instead of doubles...

cw/cw305/cw_segmented.py
```diff
@@ -89,9 +89,15 @@ def __init__(self, num_samples=740, scope_gain=23, scope=None):
...
+        if not self._is_husky:
+            offset = offset // 2
+        self._configure_scope(scope_gain, offset)
+
         self.num_segments = 1
-        self.num_samples = num_samples
+        if self._is_husky:
+            self.num_samples = num_samples
+        else:
+            self.num_samples = num_samples // 2
```
alphan, line 100: I think it would be better to have different yaml files for different setups instead of updating these numbers under the hood. WDYT?

vogelpi: I thought about that, too. But I think as most of us use Husky anyway it's probably a waste of effort to keep maintaining config files for CW-Lite also.

alphan: SG, would you mind creating an issue and adding a comment here to remind us to deprecate the cw-lite support (I guess any piece of code that checks `if husky`) when we get the chance? Thanks!

vogelpi: Good point, I've filed an issue here #67 .

alphan: Thanks!

moidx (approved):

PR #68, opened 2022-05-13
Switch to virtual Python env and install ChipWhisperer in editable mode

vogelpi:
The officially recommended way of using ChipWhisperer is to install it in editable mode. We haven't done that so far because for OpenTitan a different version is installed (due to Python 3.6) and in non-editable mode (to speed up CI).

To again use ChipWhisperer in editable mode for ot-sca as recommended and to avoid issues due to having different versions installed in non-editable and editable mode, this commit now switches to a virtual Python environment for ot-sca.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -35,3 +35,6 @@ projects/
 
 # Python requirements source directory
 src/
+
+# Python virtual environment
+.venv/
--- a/doc/getting_started.md
+++ b/doc/getting_started.md
@@ -59,13 +59,38 @@ machine or obtained using the provided [Dockerfile](https://github.com/lowRISC/o
 
 #### Installing on a Machine
 
+##### Python Virtual Environment
+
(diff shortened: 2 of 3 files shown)


.gitignore
```diff
@@ -35,3 +35,6 @@ projects/
 
 # Python requirements source directory
 src/
+
+# Python virtual environment
+venv/
```
alphan, line 40: I think a more common name for the venv directory is `.venv`.

alphan (approved):

.gitignore
vogelpi: I wasn't sure but now switched to `.venv`. Thanks for the review @alphan.

abdullahvarici (approved):
I've tested and it all works.

vogelpi:
Thanks @abdullahvarici for trying this out and feeding back. I'll merge this now.

PR #69, opened 2022-05-16
Move capture_aes_random_batch function to capture.py.

abdullahvarici:
Rename simple_capture_traces.py to capture.py. Delete
simple_capture_traces_batch.py.


abdullahvarici:
I have started with the aes_random_batch function. Instead of using scared library I've used crypto.cipher for AES encryption. Waverunner is not supported but if it should be supported, I think we can change the code to support for all capture functions in a generic way. I also have added trace range check.

cw/cw305/capture.py
```diff
@@ -200,26 +311,88 @@ def capture_aes_fvsr_key(ot):
...
+        ret = cw.capture_trace(ot.scope, ot.target, text, key, ack=False, as_int=True)
+        if not ret:
+            raise RuntimeError('Capture failed.')
+        expected = binascii.b2a_hex(cipher.encrypt(bytes(text)))
+        got = binascii.b2a_hex(ret.textout)
+        if got != expected:
+            raise RuntimeError(f'Bad ciphertext: {got} != {expected}.')
+        yield ret
+
+
```
vogelpi, line 343: This should go into a seperate commit.

cw/cw305/capture.py
```diff
@@ -200,26 +311,88 @@ def capture_aes_fvsr_key(ot):
...
 @app_capture.command()
 def aes_random(ctx: typer.Context,
-               num_traces: int = opt_num_traces,
-               plot_traces: int = opt_plot_traces):
-    """Capture AES traces from a target that runs the `aes_serial` program."""
+        num_traces: int = opt_num_traces,
+        plot_traces: int = opt_plot_traces):
+    """Capture AES traces. Fixed key random texts."""
```
vogelpi, line 348: These changes are unrelated to the batch capture work. They should go into a seperate commit. Also, it introduces again lint errors. Can you please run `flake8 capture.py --max-line-length 100` and fix all errors. Thanks!

cw/cw305/capture.py
```diff
@@ -200,26 +311,88 @@ def capture_aes_fvsr_key(ot):
...
     capture_init(ctx, num_traces, plot_traces)
     capture_loop(capture_aes_random(ctx.obj.ot, ctx.obj.ktp), ctx.obj.ot, ctx.obj.cfg["capture"])
     capture_end(ctx.obj.cfg)
 
 
+@app_capture.command()
+def aes_random_batch(ctx: typer.Context,
+        num_traces: int = opt_num_traces,
+        plot_traces: int = opt_plot_traces):
```
vogelpi, line 357: Nit: This causes lint errors.

cw/cw305/capture.py
```diff
@@ -200,26 +311,88 @@ def capture_aes_fvsr_key(ot):
...
+
+
 @app_capture.command()
 def aes_fvsr_key(ctx: typer.Context,
-                 num_traces: int = opt_num_traces,
-                 plot_traces: int = opt_plot_traces):
-    """Capture AES traces from a target that runs the `aes_serial` program."""
+        num_traces: int = opt_num_traces,
+        plot_traces: int = opt_plot_traces):
+    """Capture AES traces. TVLA DTR 5.3: fvsr keys, random texts."""
```
vogelpi, line 368: Same here: lint + it's not related to the batch capture change.

doc/getting_started.md
```diff
@@ -221,30 +221,32 @@ everything is working correctly:
 ```console
 Creating user 'ot' with UID=1000, GID=1000.
 ot@ot-sca:/repo$ cd cw/cw305/
-ot@ot-sca:/repo/cw/cw305$ ./simple_capture_traces.py 
-Connecting and loading FPGA... Done!
+ot@ot-sca:/repo/cw/cw305$ ./capture.py --cfg-file capture_aes.yaml capture aes-random
+Connecting and loading FPGA... SKIPPED! (see: /home/abdullah/workspace/ot-sca/cw/cw305/util/device.py:initialize_fpga)
```
vogelpi, line 225: Please revert this line as it assumes the FPGA has already been loaded previously. Otherwise it would just print:
```
Connecting and loading FPGA... Done!
```
as before.

doc/getting_started.md
```diff
@@ -221,30 +221,32 @@ everything is working correctly:
...
-Transferring frame 0x00000009 @ 0x00002298.
-Transferring frame 0x0000000A @ 0x00002670.
-Transferring frame 0x8000000B @ 0x00002A48.
-Serial baud rate = 38400
-Serial baud rate = 115200
-Scope setup with sampling rate 100003051.0 S/s
-Target simpleserial version: z01 (attempts: 2).
-Using key: b'2b7e151628aed2a6abf7158809cf4f3c'
+WARNING:ChipWhisperer NAEUSB:Your firmware is outdated - latest is 1.2. Suggested to update firmware, as you may experience errors
+See https://chipwhisperer.readthedocs.io/en/latest/api.html#firmware-update
```
vogelpi, line 228: Please remove these lines and update your scope firmware as described under https://chipwhisperer.readthedocs.io/en/latest/api.html#firmware-update

doc/getting_started.md
```diff
@@ -221,30 +221,32 @@ everything is working correctly:
...
+Transferring frame 0x00000005 @             0x00002738.
+Transferring frame 0x80000006 @             0x00002F10.
+Scope setup with sampling rate 200009376.0 S/s
 Reading from FPGA using simpleserial protocol.
-Checking version: 
-Capturing: 100%|████████████████████████████| 5000/5000 [00:55<00:00, 90.34it/s]
+Target simpleserial version: z01 (attempts: 1).
+Using key: b'2b7e151628aed2a6abf7158809cf4f3c'                                  
+Capturing: 100%|██████████████████████████████| 100/100 [00:01<00:00, 51.46it/s]
+Created plot with 10 traces: /home/abdullah/workspace/ot-sca/cw/cw305/projects/sample_traces_aes.html
```
vogelpi, line 249: maybe you can just replace `abdullah` here with `user` to make the guide more generic.

doc/getting_started.md
```diff
@@ -221,30 +221,32 @@ everything is working correctly:
...
+Transferring frame 0x00000004 @             0x00001F60.
+Transferring frame 0x00000005 @             0x00002738.
+Transferring frame 0x80000006 @             0x00002F10.
+Scope setup with sampling rate 200009376.0 S/s
 Reading from FPGA using simpleserial protocol.
-Checking version: 
-Capturing: 100%|████████████████████████████| 5000/5000 [00:55<00:00, 90.34it/s]
+Target simpleserial version: z01 (attempts: 1).
+Using key: b'2b7e151628aed2a6abf7158809cf4f3c'                                  
+Capturing: 100%|██████████████████████████████| 100/100 [00:01<00:00, 51.46it/s]
```
vogelpi, line 248: I think this assumes that you've changed the `num_traces` variable in the capture_aes.yaml file. Do you want to update the command above accordingly?

doc/getting_started.md
```diff
@@ -457,15 +459,12 @@ the specified file paths.
...
-* `simple_capture_traces.py`: Supports capturing AES and SHA3 (KMAC) power traces
-  using either a CW-Husky or CW-Lite capture board.
-
-* `simple_capture_traces_batch.py`: Supports capturing AES power traces
-  traces in batches to achieve substantially higher capture rates (ca.
-  1300 traces/s). Requires the CW-Husky capture board.
+* `capture.py`: Supports capturing AES and SHA3 (KMAC) power traces
+  using either a CW-Husky or CW-Lite capture board. It supports also capturing 
+  AES power traces traces in batches to achieve substantially higher capture 
+  rates (ca. 1300 traces/s). Requires the CW-Husky capture board.
```
vogelpi, line 467: `When using a CW-Husky board, the script also supports capturing AES power traces in batches to achieve substantially higher capture rates (ca. 1300 traces/s).`

doc/getting_started.md
```diff
@@ -503,13 +502,35 @@ Capturing: 100%|█████████████████████
 Created plot with 10 traces: ~/ot-sca/cw/cw305/projects/sample_traces_aes.html
 ```
 
+Following command may be used to capture AES traces in batch mode: 
+```console
+$ cd cw/cw305
+$ ./capture.py --cfg-file capture_aes.yaml capture aes-random-batch --num-traces 100 --plot-traces 10
```
vogelpi, line 508: I would select a higher number than 100 here. Say 5000.

doc/getting_started.md
```diff
@@ -503,13 +502,35 @@ Capturing: 100%|█████████████████████
...
+By default, this will capture 5000 traces which should be sufficient to make
+sure the setup is configured and working as expected before starting a long
+running capture. The plot should look very similar to the one of the
+non-batched AES capture.
+
+To capture e.g. 2 Mio traces, open the config file `capture_aes.yaml` and change
+`capture.num_traces` to `2000000`. Also, it is recommended to set
+`plot_capture.show` to `false` as the routine checking the ADC gain setting
+(as part of the plot generation) can take a long time for high number of traces.
+
```
vogelpi, line 520: Both these points are no longer required. The ADC outputs are now checked within the capture loop and independent of the plot setting. And thanks to your work, one can just provide a different `num_traces` argument using `--num-traces 2000000` :-)

doc/getting_started.md
```diff
@@ -537,64 +558,23 @@ utilization (FPGA build) will all affect the safe maximum gain setting.
...
-running capture. The plot should look very similar to the one of the
-non-batched AES capture above.
-
-To capture e.g. 2 Mio traces, open the config file `capture_aes.yaml` and change
-`capture.num_traces` to `2000000`. Also, it is recommended to set
-`plot_capture.show` to `false` as the routine checking the ADC gain setting
-(as part of the plot generation) can take a long time for high number of traces.
-
-
 ## Performing Example SCA Attack on AES with Masking Disabled
```
vogelpi, line 578: In this section can you please also remove steps 3 (set capture.num_traces) and 4 (set plot_capture.show) and instead adjust the capture command below?

doc/getting_started.md
```diff
@@ -630,7 +610,7 @@ Run the following commands
 
 ```console
 $ cd cw/cw305
-$ ./simple_capture_traces_batch.py --scope cw
+$ ./capture.py --cfg-file capture_aes.yaml capture aes-random-batch
```
vogelpi, line 613: Here we would need an argument `--num-traces 2000000`

vogelpi (commented):
Thanks @abdullahvarici for working on this! This looks already pretty good and it will simplify a lot future work using batch capture. This is cool :-)

I've added a couple of comments. It would be great if this PR could keep the focus on the batch capture. Moving the mix columns capture can go into a separate PR to simplify review. Also it would be great if we could keep the WaveRunner support. We can't test it locally but maybe @alphan will find time to do this. If not, we at least keep around the code and it will be easier in the future to re-enable WaveRunner support.

Last but not least, I would just keep using `scared` to compute the plaintexts/ciphertexts in parallel instead of switchting to the other implementation. Or this there a particular reason for doing this?

cw/cw305/capture.py
```diff
@@ -152,6 +162,107 @@ def capture_aes_random(ot, ktp):
...
+            waves = scope.capture_and_transfer_waves()
+            assert waves.shape[0] == scope.num_segments
+
+            # Generate plaintexts and ciphertexts to compare with the batch encryption results.
+            # TODO: In earlier version of the function, scared library is used for AES encryption.
+            # If it does parallel encryption, instead of using Crypto.Cipher it may be used for 
+            # better performance of the script.
+            plaintexts = [ktp.next()[1] for _ in range(scope.num_segments_actual)]
+            ciphertexts = [cipher.encrypt(bytes(plaintexts[ii]))
+                for ii in range(scope.num_segments_actual)]
```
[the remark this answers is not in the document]
abdullahvarici: In other functions in the same script we are using crypto.cipher and I thought it would be better if we use the same structure. In addition scared library is slower than crypto.cipher. When we try to capture 100k traces with scared we get 1697 traces/s and with crypto.cipher we get 1735 traces/s.

cw/cw305/capture.py
```diff
@@ -152,6 +162,107 @@ def capture_aes_random(ot, ktp):
...
+            # See:
+            # - _updateRanges() in chipwhisperer/common/api/TraceManager.py.
+            # - https://github.com/newaetech/chipwhisperer/issues/344
+            if num_segments_storage != len(project.segments):
+                if num_segments_storage >= 2:
+                    project.traces.tm.setTraceSegmentStatus(num_segments_storage - 2, False)
+                num_segments_storage = len(project.segments)
+
+            # Check that the recorded to ensure the ADC doesn't saturate.
+            check_range(waves, ot.scope.adc.bits_per_sample)
```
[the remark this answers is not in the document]
abdullahvarici: I copied to there by mistake I think. I take it back.

cw/cw305/capture.py
abdullahvarici: Oh OK.

cw/cw305/capture.py
```diff
@@ -26,9 +28,30 @@
 # Shared options for "capture aes" and "capture sha3".
 opt_num_traces = typer.Option(None, help="Number of traces to capture.")
 opt_plot_traces = typer.Option(None, help="Number of traces to plot.")
+opt_scope_type = typer.Option("cw", help=("cw or waverunner as a scope for "
+                                          "batch capture"))
```
alphan, line 32: nit: enums are safer:

```suggestion
from enum import Enum

...

class ScopeType(str, Enum):
    cw = "cw"
    waverunner = "waverunner"

opt_scope_type = typer.Option(ScopeType.cw, help=("Scope type"))
```

Please take a look at [here](https://typer.tiangolo.com/tutorial/parameter-types/enum/) for more details.

doc/getting_started.md
```diff
@@ -457,15 +450,12 @@ the specified file paths.
...
-  using either a CW-Husky or CW-Lite capture board.
+There is a one single script available for capturing OpenTitan power traces:
 
-* `simple_capture_traces_batch.py`: Supports capturing AES power traces
-  traces in batches to achieve substantially higher capture rates (ca.
-  1300 traces/s). Requires the CW-Husky capture board.
+* `capture.py`: Supports capturing AES and SHA3 (KMAC) power traces
+  using either a CW-Husky or CW-Lite capture board. When using a CW-Husky 
+  board, the script also supports capturing AES power traces in batches to 
+  achieve substantially higher capture rates (ca. 1300 traces/s).
```
alphan, line 458: nit: Since we have a single script, consider removing the list and rewording this part.

cw/cw305/capture.py
```diff
@@ -152,6 +175,102 @@ def capture_aes_random(ot, ktp):
...
+            # Transfer traces
+            waves = scope.capture_and_transfer_waves()
+            assert waves.shape[0] == scope.num_segments
+            # Check that the ADC didn't saturate when recording this batch.
+            check_range(waves, ot.scope.adc.bits_per_sample)
+
+            # Generate plaintexts and ciphertexts to compare with the batch encryption results.
+            plaintexts = [ktp.next()[1] for _ in range(scope.num_segments_actual)]
+            ciphertexts = [cipher.encrypt(bytes(plaintexts[ii]))
+                           for ii in range(scope.num_segments_actual)]
```
alphan, line 227: let's keep using `scared.aes.base.encrypt()` here.

cw/cw305/capture.py
```diff
@@ -152,6 +175,102 @@ def capture_aes_random(ot, ktp):
...
+            if num_segments_storage != len(project.segments):
+                if num_segments_storage >= 2:
+                    project.traces.tm.setTraceSegmentStatus(num_segments_storage - 2, False)
+                num_segments_storage = len(project.segments)
+
+            # Add traces of this batch to the project.
+            for wave, plaintext, ciphertext in zip(waves, plaintexts, ciphertexts):
+                project.traces.append(
+                    cw.common.traces.Trace(wave, plaintext, bytearray(ciphertext), key)
+                )
```
alphan, line 259: nit: can't remember if this was a recent change but this call is missing `dtype=np.uint16`. pointing out to avoid surprises down the road.

cw/cw305/capture.py
```diff
@@ -152,6 +175,102 @@ def capture_aes_random(ot, ktp):
...
+      capture_cfg: Capture configuration.
+      scope_type: cw or waverunner as a scope for batch capture.
+    """
+    # Seed host's PRNG.
+    # TODO: Replace this with a dedicated PRNG to avoid other packages breaking our code.
+    random.seed(capture_cfg["batch_prng_seed"])
+    # Set the target's key
+    key = ktp.next_key()
+    tqdm.write(f'Using key: {binascii.b2a_hex(bytes(key))}')
+    cipher = AES.new(bytes(key), AES.MODE_ECB)
```
alphan, line 194:
```suggestion
```

cw/cw305/capture.py
```diff
@@ -152,6 +175,102 @@ def capture_aes_random(ot, ktp):
...
+      ktp: Key and plaintext generator.
+      capture_cfg: Capture configuration.
+      scope_type: cw or waverunner as a scope for batch capture.
+    """
+    # Seed host's PRNG.
+    # TODO: Replace this with a dedicated PRNG to avoid other packages breaking our code.
+    random.seed(capture_cfg["batch_prng_seed"])
+    # Set the target's key
+    key = ktp.next_key()
+    tqdm.write(f'Using key: {binascii.b2a_hex(bytes(key))}')
```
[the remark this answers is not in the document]
abdullahvarici: For checking, I moved `random.seed()` before `capture_init()` and saw that the same sequence of key and pt is generated.

cw/cw305/capture.py
abdullahvarici: Is there a specific reason for that? In other functions in the same script we are using crypto.cipher and I thought it would be better if we use the same structure. In addition scared library is slower than crypto.cipher. When we try to capture 100k traces with scared we get 1697 traces/s and with crypto.cipher we get 1735 traces/s.

cw/cw305/capture.py
abdullahvarici: When I try to add `np.uint16` as an argument to `cw.common.traces.Trace()` function, it gives an error saying "takes 5 positional arguments but 6 were given".

cw/cw305/capture.py
abdullahvarici: Should we use scared function instead?

abdullahvarici:
Thanks @alphan for the review. I have already applied and saw that different capture functions works with different parameters. I have made changes according to your review and forced-pushed them.

cw/cw305/capture.py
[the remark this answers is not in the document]
vogelpi: Thanks.

cw/cw305/capture.py
vogelpi: Thanks.

doc/getting_started.md
vogelpi: Thanks.

doc/getting_started.md
vogelpi: Actually, you now wrote `--num-traces=2000000` does that work too or do we need `--num-traces 2000000` (without equal sign)?

cw/cw305/capture.py
vogelpi: I don't remember to be honest. But in general I would just change one thing at a time (merge batch script into normal script for this PR). The difference in performance you see is roughly 2%. It's not too much but maybe also not negligible. This change could be another PR. Up to now, two people have reviewed this PR and this point has been brought up by both. So please move the change into a separate PR to unblock this one.

cw/cw305/capture.py
vogelpi: The `dtype=np.uint16` argument is required for the `append()` function not the `Trace()`. Without this argument, the trace files will again be 4x bigger. This has been added very recently and it probably got forgotten when you rebased your work. 

This is a good example why we don't want to change more than one thing at once/per PR. Please read this section from the contribution guide to learn how we structure PRs and commits: https://docs.opentitan.org/doc/project/detailed_contribution_guide/#working-with-pull-requests

cw/cw305/capture.py
vogelpi: No, let's not do that now. We can file a separate PR to use the same cipher everywhere. This PR should just be about merging the batch script into the regular one.

cw/cw305/capture.py
vogelpi: Good catch @alphan by the way!

doc/getting_started.md
abdullahvarici: My mistake :) Changed it.

vogelpi (approved):
Thanks a lot @abdullahvarici for implementing the remaining feedback! Maybe we wait for a couple more hours before merging in case @alphan want to take another look. But from my side, this looks definitely good!

cw/cw305/capture.py
vogelpi: Cool, I've tried out the PR. 100000 traces are ~235 MiB in size. So everything as expected ;-)

alphan:
> Thanks a lot @abdullahvarici for implementing the remaining feedback! Maybe we wait for a couple more hours before merging in case @alphan want to take another look. But from my side, this looks definitely good!

Apologies for the delay, LGTM, thanks @abdullahvarici!

PR #70, opened 2022-05-20
Move aes_mix_columns_capture function into capture.py

abdullahvarici:
There are two commits. The first one is actually from PR69.

At this commit, aes_mix_columns_capture function is moved into capture.py. 6000 traces is captured using an unmasked implementation and analysed with mix_columns_cpa_attack.py script. All 16 bytes of the key are found successfully. Getting started guide is also updated.


cw/cw305/capture.py
```diff
@@ -361,6 +361,34 @@ def aes_fvsr_key(ctx: typer.Context,
...
 
 
+@app_capture.command()
+def aes_mix_column(ctx: typer.Context,
+                   num_traces: int = opt_num_traces,
+                   plot_traces: int = opt_plot_traces):
+    """Capture AES traces. Fixed key, Random texts. 4 sets of traces. Mix Column HD CPA Attack.
+    Attack implemented by ChipWhisperer:
+    Repo: https://github.com/newaetech/chipwhisperer-jupyter/blob/master/experiments/...
+    ...MixColumn%20Attack.ipynb
```
vogelpi, line 371: Here you don't need to wrap the line as it breaks the URL. You can add the following tag at the very end of the line to not make the lint tool complain: `# noqa: E501`

vogelpi (commented):
This looks good @abdullahvarici . I didn't try it out on the FPGA yet though.

You can now rebase this on master as I've merged PR #69. Thanks!

vogelpi (approved):
LGTM!

cw/cw305/capture.py
```diff
@@ -361,6 +361,33 @@ def aes_fvsr_key(ctx: typer.Context,
...
+                   plot_traces: int = opt_plot_traces):
+    """Capture AES traces. Fixed key, Random texts. 4 sets of traces. Mix Column HD CPA Attack.
+    Attack implemented by ChipWhisperer:
+    Repo: https://github.com/newaetech/chipwhisperer-jupyter/blob/master/experiments/MixColumn%20Attack.ipynb # noqa: E501
+    Reference: https://eprint.iacr.org/2019/343.pdf
+    See mix_columns_cpa_attack.py for attack portion.
+    """
+    capture_init(ctx, num_traces, plot_traces)
+
+    ctx.obj.ktp = cw.ktp.VarVec()
```
alphan, line 376: why are the lines that set `key_len` and `text_len` are dropped here?

alphan:
Thanks @abdullahvarici!

cw/cw305/capture.py
abdullahvarici: Thanks Alphan. I've dropped them by mistake. Should I add them with another PR?

vogelpi: Yes, I think that would be the best.

abdullahvarici: Done it: https://github.com/lowRISC/ot-sca/pull/73

PR #73, opened 2022-05-24
Add lines to set key and text len in AES mix column capture.

abdullahvarici:


alphan (approved):

vogelpi (approved):
LGTM!

PR #74, opened 2022-05-26
Add support for AES TVLA fixed-vs-random key captures in batch mode.

abdullahvarici:
In order to simplify the analysis, the first encryption has to use
fixed key. In addition a PRNG is used for random key and plaintext
generation instead of AES algorithm as specified in the TVLA DTR.
aes_serial firmware is also updated to support this functionality.

### DIFF
--- a/cw/cw305/capture.py
+++ b/cw/cw305/capture.py
@@ -330,6 +330,124 @@ def capture_aes_fvsr_key(ot):
         yield ret
 
 
+def capture_aes_fvsr_key_batch(ot, ktp, capture_cfg, scope_type):
+    """A generator for capturing AES traces for fixed vs random key test in batch mode.
+    The data collection method is based on the derived test requirements (DTR) for TVLA:
+    https://www.rambus.com/wp-content/uploads/2015/08/TVLA-DTR-with-AES.pdf
+    The measurements are taken by using either fixed or randomly selected key in batches.
+    In order to simplify the analysis, the first encryption has to use fixed key.
+    To generate random keys and texts internal PRNG's are used instead of AES functions as specified
+    in the DTR.
+
+    Args:
+      ot: Initialized OpenTitan target.
+      ktp: Key and plaintext generator.
+      capture_cfg: Capture configuration.
+      scope_type: cw or waverunner as a scope for batch capture.
+    """
+    # Seed host's PRNG.
+    # TODO: Replace this with a dedicated PRNG to avoid other packages breaking our code.
+    random.seed(capture_cfg["batch_prng_seed"])
+    # Seed the target's PRNG
+    ot.target.simpleserial_write("s", capture_cfg["batch_prng_seed"].to_bytes(4, "little"))
+
+    key_fixed = bytearray([0x81, 0x1E, 0x37, 0x31, 0xB0, 0x12, 0x0A, 0x78,
+                           0x42, 0x78, 0x1E, 0x22, 0xB2, 0x5C, 0xDD, 0xF9])
+    tqdm.write(f'Fixed key: {binascii.b2a_hex(bytes(key_fixed))}')
+    ot.target.simpleserial_write("t", key_fixed)
+
+    sample_fixed = 1
+
+    # Create the ChipWhisperer project.
+    project = cw.create_project(capture_cfg["project_name"], overwrite=True)
+    # Capture traces.
+    rem_num_traces = capture_cfg["num_traces"]
+    num_segments_storage = 1
+    # cw and waverunner scopes are supported for batch capture.
+    scope = SCOPE_FACTORY[scope_type](ot, capture_cfg)
+
+    with tqdm(total=rem_num_traces, desc="Capturing", ncols=80, unit=" traces") as pbar:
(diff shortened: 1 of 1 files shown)


abdullahvarici:
Please see also https://github.com/lowRISC/opentitan/pull/12886.

alphan (approved):
Thanks @abdullahvarici! lgtm but @vogelpi should also review this.

I think we will also need to update the aes_serial binary once lowRISC/opentitan#2886 is merged.

cw/cw305/capture.py
```diff
@@ -330,6 +330,124 @@ def capture_aes_fvsr_key(ot):
...
+            plaintexts = []
+            ciphertexts = []
+            keys = []
+            for ii in range(scope.num_segments_actual):
+                if sample_fixed:
+                    key = np.asarray(key_fixed)
+                else:
+                    key = np.asarray(ktp.next()[1])
+                plaintext = np.asarray(ktp.next()[1])
+                ciphertext = np.asarray(scared.aes.base.encrypt(plaintext, key))
```
vogelpi, line 394: I think you found out in the meantime that computing all ciphertexts here has a very high impact on the capture rate. On the target we're no longer concatenating the encryptions so we should probably also remove it here?

vogelpi: Actually I am not sure if the high impact on the capture rate is justified. Because we have also been computing all ciphertexts in `capture_aes_random_batch()` and there we have a high capture rate as well. Is the problem maybe in how this piece of code here is written? Maybe too many copies?

cw/cw305/capture.py
```diff
@@ -330,6 +330,124 @@ def capture_aes_fvsr_key(ot):
...
+            # Add traces of this batch to the project.
+            for wave, plaintext, ciphertext, key in zip(waves, plaintexts, ciphertexts, keys):
+                project.traces.append(
+                    cw.common.traces.Trace(wave, plaintext, ciphertext, key),
+                    dtype=np.uint16
+                )
+
+            # Update the loop variable and the progress bar.
+            rem_num_traces -= scope.num_segments
+            pbar.update(scope.num_segments)
```
vogelpi, line 440: This code is exactly the same as for `capture_aes_random_batch()` does it make sense to share it by creating functions?

cw/cw305/capture.py
```diff
@@ -330,6 +330,124 @@ def capture_aes_fvsr_key(ot):
...
+            scope.arm()
+
+            # Start batch encryption.
+            ot.target.simpleserial_write("f", scope.num_segments_actual.to_bytes(4, "little"))
+
+            # Transfer traces.
+            waves = scope.capture_and_transfer_waves()
+            assert waves.shape[0] == scope.num_segments
+            # Check that the ADC didn't saturate when recording this batch.
+            check_range(waves, ot.scope.adc.bits_per_sample)
```
vogelpi, line 382: Does it have any impact on the capture rate if the reading out/transferring of traces is moved down to after computing the ciphertexts? Because while the encryption is running, the host could theoretically do something else instead of waiting for the capture to finish in `scope.capture_and_transfer_waves()`.

vogelpi (commented):
I have had another look at this. It mostly looks good but I have some question regarding re-ordering/optimizing code blocks to improve capture performance and sharing code.

cw/cw305/capture.py
abdullahvarici: I was thinking that I can fix this in another PR with some other fixes. In `capture_aes_random_batch()` we were using the same key for all encyrptions. But in `capture_aes_fvsr_key_batch()` we are setting new key for almost every encryption. This may be the reason for that.

cw/cw305/capture.py
abdullahvarici: I was thinking the same. What would be better: In the same PR or in a seperate PR?

cw/cw305/capture.py
abdullahvarici: Yes it also increased the capture rate. I was going to add that in a seperate PR.

cw/cw305/capture.py
m-temp: Maybe these are two very stupid questions/comments but:
1)  why do you assign `np.asarray(ktp.next()[1])` to both key and plaintext?  Is there a typo for the index?
2) are you aware that  `ktp.next` might produce the same key all the time? https://forum.newae.com/t/scope-function-for-cwnano-correlation-power-analysis/2806/8 defined in `software/chipwhisperer/capture/acq_patterns/basic.py`

abdullahvarici: When I call `ktp.next()[1]` over and over again, it generates random number each time. I am basically calling it to generate random numbers.

m-temp: yeah. I just realized that, but you were faster than my edit ;)  Thanks!

vogelpi: Investigating this later on sounds good to me.

cw/cw305/capture.py
vogelpi: Sounds good to me.

PR #77, opened 2022-06-06
Optimize and clean-up AES capture functions.

abdullahvarici:
Capture rate increased from 300 traces/s to 880 traces/s for
AES fixed-vs-random key batch capture. Also some cleanup is
done.

### DIFF
--- a/cw/cw305/capture.py
+++ b/cw/cw305/capture.py
@@ -37,15 +37,16 @@ app.add_typer(app_capture, name="capture", help="Capture traces for SCA")
 opt_num_traces = typer.Option(None, help="Number of traces to capture.")
 opt_plot_traces = typer.Option(None, help="Number of traces to plot.")
 opt_scope_type = typer.Option(ScopeType.cw, help=("Scope type"))
+opt_ciphertexts_store = typer.Option(False, help=("Store all ciphertexts for batch capture."))
 
 
 def create_waverunner(ot, capture_cfg):
-    """Create a WaveRunner object to capture traces faster"""
+    """Create a WaveRunner object to be used for batch capture."""
     return WaveRunner(capture_cfg["waverunner_ip"])
 
 
 def create_cw_segmented(ot, capture_cfg):
-    """Create CwSegmented object to capture traces faster."""
+    """Create CwSegmented object to be used for batch capture."""
     return CwSegmented(num_samples=capture_cfg["num_samples"],
                        offset=capture_cfg["offset"],
                        scope_gain=capture_cfg["scope_gain"],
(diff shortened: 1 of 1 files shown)


abdullahvarici:
Any more recommendation for improving capture performance and cleanup is appreciated. Please see https://github.com/lowRISC/opentitan/pull/13057.

cw/cw305/capture.py
```diff
@@ -357,6 +395,7 @@ def capture_aes_fvsr_key_batch(ot, ktp, capture_cfg, scope_type):
     ot.target.simpleserial_write("t", key_fixed)
 
     sample_fixed = 1
+    first_batch = 1
```
vogelpi, line 398: This should probably be a `bool` using `True/False`.

cw/cw305/capture.py
```diff
@@ -227,45 +269,20 @@ def capture_aes_random_batch(ot, ktp, capture_cfg, scope_type):
...
-            ]
-            # Check the last ciphertext of this batch to make sure we are in sync.
-            actual_last_ciphertext = ot.target.simpleserial_read("r", 16, ack=False)
-            expected_last_ciphertext = ciphertexts[-1]
-            assert actual_last_ciphertext == expected_last_ciphertext, (
-                f"Incorrect encryption result!\n"
-                f"actual: {actual_last_ciphertext}\n"
-                f"expected: {expected_last_ciphertext}"
+            expected_last_ciphertext = np.asarray(scared.aes.base.encrypt(
+                np.asarray(plaintexts[-1]), np.asarray(key))
```
vogelpi, line 275: We actually need this for some attack scripts (e.g. `ceca.py` can be applied on the outputs for which the ciphertexts of all blocks are needed) and if I remember correctly as long as the key remeins the same, the performance impact isn't huge.

So, the PR should be changed to disable the ciphertext generation on the host only for `aes_fvsr_batch()` but still allow to enable that it turns out we need it.

vogelpi (commented):
Thanks for your PR @abdullahvarici, the improvement in terms of capture rate is very impressive !

This looks mostly good. I think for `aes_random_batch()` we need to keep generating the ciphertext (because some attacks need it), and for the `aes_fvrs_key_batch()` we should also have a switch to enable it.

cw/cw305/capture.py
```diff
@@ -372,66 +411,50 @@ def capture_aes_fvsr_key_batch(ot, ktp, capture_cfg, scope_type):
             scope.num_segments = min(rem_num_traces, scope.num_segments_max)
             scope.arm()
 
-            # Start batch encryption.
+            # Start batch encryption. In order to increase capture rate, after the first batch
```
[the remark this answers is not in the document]
abdullahvarici: Using both optimization increases capture rate to 880 traces/s. But if we don't use this optimization (i.e. we do not generate ciphertexts as an optimization) 659 trace/s.

cw/cw305/capture.py
```diff
@@ -37,15 +37,16 @@ class ScopeType(str, Enum):
 opt_num_traces = typer.Option(None, help="Number of traces to capture.")
 opt_plot_traces = typer.Option(None, help="Number of traces to plot.")
 opt_scope_type = typer.Option(ScopeType.cw, help=("Scope type"))
+opt_ciphertexts_gen = typer.Option(False, help=("Generate all ciphertexts for batch capture."))
```
vogelpi, line 40: This would better be called `opt_ciphertexts_store` because the ciphertexts are generated anyway on the target. It's the result of the encryption.

vogelpi: Then, where the argument is used inside `aes_fvsr_key_batch()` there should be commend saying by default we don't store the ciphertexts as generating them on the host as well as transferring them over from the target substantially reduces capture performance. It should therefore only be enabled if absolutely needed.

vogelpi (commented):
I had another look and now some more comments. Please take a look.

cw/cw305/capture.py
```diff
@@ -235,37 +281,15 @@ def capture_aes_random_batch(ot, ktp, capture_cfg, scope_type):
...
-                num_segments_storage = len(project.segments)
+            check_ciphertext(ot, ciphertexts[-1], True)
+
+            num_segments_storage = optimize_cw_capture(project, num_segments_storage)
 
             # Add traces of this batch to the project.
             for wave, plaintext, ciphertext in zip(waves, plaintexts, ciphertexts):
                 project.traces.append(
-                    cw.common.traces.Trace(wave, plaintext, bytearray(ciphertext), key),
+                    cw.common.traces.Trace(wave, plaintext, ciphertext, key),
```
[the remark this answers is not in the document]
abdullahvarici: Yes, that's right.

cw/cw305/capture.py
```diff
@@ -372,68 +418,66 @@ def capture_aes_fvsr_key_batch(ot, ktp, capture_cfg, scope_type):
...
+            # encryption, send "g" to the device to generate random keys and plaintexts at the
+            # device when this script is getting waves from the scope.
+            if is_first_batch:
+                ot.target.simpleserial_write("g", scope.num_segments_actual.to_bytes(4, "little"))
+                is_first_batch = False
             ot.target.simpleserial_write("f", scope.num_segments_actual.to_bytes(4, "little"))
+            next_batch_size = min(rem_num_traces - scope.num_segments_actual,
+                                  scope.num_segments_max)
+            if next_batch_size > 0:
+                ot.target.simpleserial_write("g", next_batch_size.to_bytes(4, "little"))
```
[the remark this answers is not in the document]
abdullahvarici: Thanks @vogelpi. This worked and I also thought that it would increase the capture rate too since we are using serial communication less this way but it didn't change at all. Maybe we have a bottleneck somewhere else.

vrozic (approved):
I've tested this change on CW310 and the capturing works correctly.

cw/cw305/capture.py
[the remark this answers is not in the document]
vogelpi: Sounds good, thanks.

cw/cw305/capture.py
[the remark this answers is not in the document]
vogelpi: Thanks for changing it. I think the code is more reliable like this. Don't worry about the capture rate.

cw/cw305/capture.py
vogelpi: Perfect, thanks!

cw/cw305/capture.py
```diff
@@ -37,15 +37,16 @@ class ScopeType(str, Enum):
 opt_num_traces = typer.Option(None, help="Number of traces to capture.")
 opt_plot_traces = typer.Option(None, help="Number of traces to plot.")
 opt_scope_type = typer.Option(ScopeType.cw, help=("Scope type"))
+opt_ciphertexts_store = typer.Option(False, help=("Generate all ciphertexts for batch capture."))
```
vogelpi, line 40: Nit: also the comment/help text should be adjusted accordingly.

cw/cw305/capture.py
```diff
@@ -372,7 +418,12 @@ def capture_aes_fvsr_key_batch(ot, ktp, capture_cfg, scope_type):
             scope.num_segments = min(rem_num_traces, scope.num_segments_max)
             scope.arm()
 
-            # Start batch encryption.
+            # Start batch encryption. In order to increase capture rate, after the first batch
+            # encryption, the device will start automatically to generate random keys and plaintexts
+            # when this script is getting waves from the scope.
+            if is_first_batch:
+                ot.target.simpleserial_write("g", scope.num_segments_actual.to_bytes(4, "little"))
```
vogelpi, line 425: I think you can just move this first call to `ot.target.simpleserial_write("g", scope.num_segments_actual.to_bytes(4, "little"))` to the outside of the loop. Then you can also remove the the `is_first_batch` variable completely.

vogelpi (approved):
I got two more minor comments. Feel free to merge the PR once these have been addressed.

About updating the firmware binaries: I can do that for you on THU or FRI and in a separate PR. I can do it and test it for both boards at the same time. Ideally until then we also have the KMAC batch capture merged. You can merge this PR after addressing the other two comments. Thanks!

abdullahvarici:
Hi @vogelpi. I have edited the PR according to your last two comments but I am not sure how to merge this PR.

PR #79, opened 2022-06-16
Fix a bug in AES FvsR key batch capture function.

abdullahvarici:
We need to compute scope.num_segments at first.


vogelpi (approved):
LGTM!

PR #80, opened 2022-06-16
Update binaries and getting started guide.

abdullahvarici:
Update binaries for AES FvsR key batch capture. Update getting
started guide especially with bazel commands.

### DIFF
--- a/doc/getting_started.md
+++ b/doc/getting_started.md
@@ -263,8 +263,7 @@ OpenTitan using the CW310, follow these steps:
 1. Build the regular binaries with
 ```console
 $ cd $REPO_TOP
-$ ./meson_init.sh
-$ ninja -C build-out all
+$ ./bazelisk.sh build //sw/...
 ```
 1. Then, the bitstream generation can be started by running
 ```console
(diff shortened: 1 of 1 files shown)


vogelpi (changes requested):
Thanks for the PR! The changes to the documentation look mostly good, but they should go into separate commit at least. Because they are not related to updating the binaries.

I've tested the binaries on both boards. The `aes-random(-batch)` commands work fine (previously broken) but `aes-fvsr-key(-batch)` doesn't work on either boards. I get the following error:
```
AssertionError: Incorrect encryption result!
actual: bytearray(b'^\xdc\xe8{')
expected: [ 85 154 226   3  30  12  31 147  54 197 151  35   4 134  80  98]
```
Can you please investigate this issue and fix it?

abdullahvarici:
Thanks @vogelpi for the review. I have updated the guide and force-pushed it. I've also created a new PR for the binaries as you suggested. In my side, I tested it on CW310 again and saw that it is working. (I also cloned it to another destination and checked-out to the PR and saw that it is working.) Maybe someone else can try them too, @vrozic?

vogelpi (approved):
LGTM, thanks for factoring out the binary updates.

PR #81, opened 2022-06-17
Update ChipWhisperer

vogelpi:
This commit updates ChipWhisperer. As a result, the warnings related to the ADC frequency exceeding the specification (200 MHz) go away. Related to this update, the firwmare of Husky should be updated (follow official instructions).


vogelpi:
After applying this commit, you need to source the virtual environment and then type:
```
pip3 install --upgrade -r python-requirements.txt
```

alphan (approved):

PR #82, opened 2022-06-17
Update binaries for AES FvsR key batch capture.

abdullahvarici:


vogelpi (changes requested):
AES fvsr key batch capture is still broken with these binaries. I keep getting:
```
AssertionError: Incorrect encryption result!
actual: bytearray(b'f\xe9K\xd4')
expected: [202  63  85  27  96  86 213 173  25 192 235 167 113 104 159 138]
```
I've tested different combinations and rebooted the FPGA (to make sure to get the latest bitstream as well) but it doesn't help. I think the issue is related to how we do the assertion: one is a string the other one a byte array. We're missing a bytearray conversion.

Can you please provide the exact command you're running as well as the Python version you are using?

abdullahvarici:
My python version is 3.8.10. I am using this command to capture:
`./capture.py --cfg-file capture_aes_cw310.yaml capture aes-fvsr-key-batch --num-traces=1000 --plot-traces=100`

vrozic (approved):
I've tested the changes on CW310. It all works correctly.

andreaskurth (commented):
Thanks, @abdullahvarici!

The updated binaries work for me on the CW310:
```sh
cd cw/cw305
./capture.py --cfg-file capture_aes_cw310.yaml capture aes-fvsr-key-batch --num-traces=1000 --plot-traces=100
```
results in
bokeh_plot(2)

The first commit in this PR seems to be completely unrelated to the binary update. I suggest to put the first commit into a separate PR.

abdullahvarici:
Thanks @andreaskurth,
The first commit belongs the previous PR actually and it is merged already. Now I am rebasing from there.

abdullahvarici:
By the way, have you updated the Chipwhisperer firmware as explained in https://github.com/lowRISC/ot-sca/pull/81 @andreaskurth and @vogelpi? Because I and @vrozic didn't updated it yet. If the maximum number of traces that the scope can capture is changed after update, it may lead the problem that you are encountering @vogelpi.

vogelpi:
That's a good point. I did indeed update the scope firmware. However, the max number of traces didn't change. After also checking out a fresh copy of the repo and digging some more it seems there is a timing issue. If I for example add a sleep here and there or add the --gen-ciphertext argument things work, although not reliable. Sometimes I get a failure in the first batch, sometimes in a later batch, sometimes it just works.

vogelpi (approved):
As I am the only one seeing this failure at the moment it's probably best to just merge the PR. I've opened #83 to track the outstanding issue.

PR #85, opened 2022-06-28
Vendor in lintpy.py from OpenTitan

andreaskurth:
- Add linters to Python requirements.
- Vendor in `lintpy.py` and style files for flake8 and yapf from
  `lowRISC/opentitan@8e0201f02d` under `util/vendor/lowrisc_opentitan`.
- Add symlinks to the style files in the root of the repository, so the
  linters can find them.


vogelpi (approved):
LGTM!

PR #91, opened 2022-06-28
Remove cw305 directory level

andreaskurth:
This moves the files under `cw/cw305/` into `cw/`.  The intermediary
`cw305` directory was misleading because that directory contained code
also for the CW310.


vogelpi (commented):
Thanks for doing this @andreaskurth .

I am not sure what happens to the binaries and bitstreams exactly when moving them as they come from Git LFS. It could be that they are added to Git with this PR. I think you probably also have to update `.gitattributes` accordingly. This file can be found in the top-level directory.

andreaskurth:
Thanks for your feedback, @vogelpi. I had forgotten the hidden files. Now `.gitattributes` and `.gitignore` have been updated, too.

The Git-internal objects show that the files have simply been *renamed*, including the binaries. This means all LFS-related data remains intact and the binaries content did and does not get directly added to the repository.

andreaskurth:
We should run a capture as sanity check

vogelpi (approved):
I just ran a test capture + tvla on the CW310. It all works fine, LGTM!

PR #92, opened 2022-06-28
Add typer CLI and yaml file support to tvla script.

abdullahvarici:
Default configurations are specified in yaml file and
it is possible to overwrite them over command line interface.
This is provided by typer library. It also provides descriptions
when typed --help. Updated configurations are written into
a .yaml file after analysis.

### DIFF
--- a/cw/tvla.py
+++ b/cw/tvla.py
@@ -2,35 +2,27 @@
 # Copyright lowRISC contributors.
 # Licensed under the Apache License, Version 2.0, see LICENSE for details.
 # SPDX-License-Identifier: Apache-2.0
-"""Test Vector Leakage Assessment
 
-
-
-Typical usage:
-
-To run the analysis without loading or saving the histograms:
->>> ./tvla.py
-
-To save histograms in the OUTPUT_FILE for later use:
->>> ./tvla.py -o OUTPUT_FILE
-
-To load histograms from the INPUT_FILE
->>> ./tvla.py -i INPUT_FILE
-
-"""
-
-import argparse
+import inspect
 import logging as log
 import multiprocessing
 import os
 from pathlib import Path
+from types import SimpleNamespace
 
 import chipwhisperer as cw
 import matplotlib.pyplot as plt
 import numpy as np
+import typer
+import yaml
 from chipwhisperer.analyzer import aes_funcs
 from joblib import Parallel, delayed
 from scipy.stats import ttest_ind_from_stats
+
+app = typer.Typer(add_completion=False)
+
+
+script_dir = Path(__file__).parent.absolute()
 
 
 class UnformattedLog(object):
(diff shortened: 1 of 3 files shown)


abdullahvarici:
Can you test the entire functionality? Because I don't know with which configurations we should test it.

abdullahvarici:
Fixed the issue when overwriting a configuration from True to False and force-pushed.

abdullahvarici:
CLI is changed a bit. New usage:
`./tvla_run.py [options] run-tvla`

If the cfg file is not provided, it will use the default options (you can see them by typing `./tvla_run.py --help`) and if you provide `--update-cfg-file` option it will either update cfg-file or create a new one if it is not provided.

cw/cw305/tvla_aes.yaml
```diff
@@ -0,0 +1,15 @@
...
+number_of_steps: 10
+output_file: null
+plot_figures: false
+project_file: projects/opentitan_simple_aes.cwp
+round_select: null
+save_to_disk: true
+trace_end: null
+trace_file: null
+trace_start: null
+ttest_step_file: null
```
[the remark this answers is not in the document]
abdullahvarici: Yes, we can create and use yaml files but there is no need for this file to be in the repository.

andreaskurth:
I considered this in more detail, and I think it does not yet do quite what we want -- or at least the code could be streamlined to implement what we want. I could be wrong, but let's clarify this.

First, do we agree that argument/parameter values should be determined by the following flow:
1. In the Python code, we specify a sane default value for each argument. That default value is defined once and it appears in the help message so it is transparent to the user.
2. When run, the TVLA script tries to load a config file. The path of that config file is given by the command-line argument or the default specified in the code.
2.1. If the config file does not exist and the path comes from the default value, nothing happens.
2.2. If the config file does not exist and the path comes from a command-line argument, an error is raised and execution is aborted.
2.3. If the config file exists, its values are loaded. Any values defined in the configuration file override the defaults specified in the code (except for the path to the config file, which has already been decided).
3. Command-line arguments have the "final word" and override any values that have been decided before.

What do you think?

(From how I understand the code, I think it does *almost* that but not exactly.)

andreaskurth (commented):
The definition of the default values and the help text now LGTM; thanks @abdullahvarici!

As discussed yesterday, it would be great if you could extent `test/tvla_test.py` to check different cases of parameter passing (defaults from code, loaded from a config file, overridden by command-line arguments). Also, we should not forget to test this with actual measurements.

(I don't mean this in a "pushy" way, just wanted to capture here what we discussed yesterday regarding this PR. :slightly_smiling_face: )

vogelpi:
I've forgot a bit about this. I think we want should verify correct functionality of the following argument combinations before merging this:

AES specific:
1. Single step, save intermediate results and histograms: `-d -o OUTPUT_FILE`
2. Single step, load trace file: `-t TRACE_FILE`
3. Single step, load trace and leakage file: `-t TRACE_FILE -l LEAKAGE_FILE`
4. Single step, load trace, leakage and t-test file: `-t TRACE_FILE -l LEAKAGE_FILE -a TTEST_STEP_FILE`
5. Single step, load trace, leakage and histograms  `-t TRACE_FILE -l LEAKAGE_FILE -i INPUT_FILE`
6. Stepped, save intermediate results: `-n 10 -d`
7. Stepped, load t-test file and produce figures only `-n 10 -a TTEST_FILE -f`
8. Stepped, load intermedaite results, round 10 only: `-n 10 -a TTEST_FILE -f -r 10`
9. Stepped, save intermediate results, round 10 only: `-n 10 -d -r 10`
10. Stepped, load intermedaite results, round 10 only: `-n 10 -a TTEST_FILE -f -r 10` (same as above but also check if all rounds were saved but just one is loaded)
11. Stepped, load intermedaite results, round 10 only and byte 9 only: `-n 10 -a TTEST_FILE -f -r 10 -b 9`
12. Stepped, save intermediate results, round 10 only and byte 9 only: `-n 10 -d -r 10 -b 9`
13. Stepped, load intermedaite results, round 10 only and byte 9 only: `-n 10 -a TTEST_FILE -f -r 10 -b 9` (same as above but also check if all bytes were saved but just one is loaded)

AES general:
1. Stepped, save intermediate results: `-n 10 -d -g -m aes`
2. Stepped, load t-test file and produce figures only `-n 10 -a TTEST_FILE -f -g -m aes`
3. Stepped, save intermediate results, use subset of traces only, produce figures: `-n 10 -d -g -m aes -f -s 1000 -e 100000` (may need to change numbers to something more reasonable)

KMAC general
1. Stepped, save intermediate results: `-n 10 -d -g -m sha3`
2. Stepped, load t-test file and produce figures only `-n 10 -a TTEST_FILE -f -g -m sha3`

Are there others? Please add them to the list.

abdullahvarici:
I have rebased and force-pushed the PR. @vrozic is testing various combinations which @vogelpi mentioned above. And I think it would be a good idea to create a seperate PR to extent `test/tvla_test.py` to check different cases since this PR is already big enough. Would it be OK for you @andreaskurth?

andreaskurth:
Yes, I agree with merging this as soon as the correct functionality of the cases specified by @vogelpi has been confirmed (manually). We can then create separate PRs to automatically test those cases in CI

cw/tvla.py
```diff
@@ -787,18 +668,18 @@ def main():
...
                 histograms_in = histograms
 
             # Histograms can be saved for later use if output file name is passed.
-            if args.output_file is not None:
+            if cfg["output_file"] is not None:
                 log.info("Saving Histograms")
-                np.savez(args.output_file, histograms=histograms, rnd_list=rnd_list,
-                         byte_list=byte_list, single_trace = traces[1])
+                np.savez(cfg["output_file"], histograms=histograms, rnd_list=rnd_list,
+                         byte_list=byte_list)
```
vrozic, line 682: It looks like `single_trace = traces [1]` was removed from this line by accident. Could you return it?

abdullahvarici: I returned it. Thanks

vrozic (approved):
Thanks @abdullahvarici 
This looks good to me.
I've tested all the cases suggested by @vogelpi and everything that was working before is still working.

For AES specific test 5. (load trace file, leakage and histograms) fails but this is due to the pre-existing issue.

I suggest merging these changes now and resolving the remaining issues in a separate PR.

PR #93, opened 2022-06-30
pytest: Add utilities to test CLI programs, add a first TVLA test, and add GitHub workflow

andreaskurth:


vogelpi (approved):
This looks good, thanks @andreaskurth !

Eventually, we will run more TVLA cases and will need more compute. It will then make sense to move this into an Azure runner. But for now, this is good. It's a big improvement to the infrastructure.

PR #94, opened 2022-06-30
Trigger lintpy workflow on pull requests

andreaskurth:
Without the `pull_request` event, the workflow would only trigger on
pushes to branches in the repository.  When pushing to a forked branch,
the workflow would not trigger.  Thus, the `pull_request` event is
required to trigger the workflow in PRs.  The `push` event restricted to
the `master` branch additionally covers pushes directly to the master
branch, without creating a duplicate check on pushes to pull request
branches.

### DIFF
--- a/.github/workflows/lintpy.yml
+++ b/.github/workflows/lintpy.yml
@@ -4,8 +4,11 @@ name: lintpy
 
 # Controls when the workflow will run
 on:
-  # Triggers the workflow on push events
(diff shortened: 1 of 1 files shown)


vogelpi (approved):
Thanks for the PR @andreaskurth.

I thought to have observed that with the previous setting pushes to PRs also triggered the workflow. But I see we use the exact same triggers you now have in your PR also for other repositories. LGTM!

PR #89, opened 2022-06-28
[tvla] Enable histogram loading for general tests.

vrozic:
Signed-off-by: Vladimir Rozic <vrozic@lowrisc.org>

PR #95, opened 2022-06-30
TVLA: Test histogram loading for general tests

andreaskurth:
This tests the functionality implemented in #89.

### DIFF
--- a/test/tvla_test.py
+++ b/test/tvla_test.py
@@ -2,8 +2,10 @@
 # Licensed under the Apache License, Version 2.0, see LICENSE for details.
 # SPDX-License-Identifier: Apache-2.0
 
+import numpy as np
+
 from .cmd import Args
-from .repo import RepoCmd
+from .repo import RepoCmd, TestDataPath
 
 
 class TvlaCmd(RepoCmd):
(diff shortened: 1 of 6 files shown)


test/tvla_test.py
```diff
@@ -17,3 +19,24 @@ def test_help():
     tvla = TvlaCmd(Args('--help')).run()
     # Assert that a message is printed on stdout or stderr.
     assert(len(tvla.stdout()) != 0 or len(tvla.stderr()) != 0)
+
+
+def ttest_significant(ttest_trace) -> bool:
+    """Determine if a t-test trace contains a significant deviation from the mean."""
+    mean = np.mean(ttest_trace, axis=3)
+    stddev = np.std(ttest_trace, axis=3)
+    threshold = mean + 4.5 * stddev
```
vrozic, line 28: This should be replaced with just
`threshold = 4.5`

The data stored in `ttest.npy` are the t-test results, so we only want to check if any of the values is outside of the (-4.5, 4.5) interval

andreaskurth: Thanks, @vrozic! Fixed.

andreaskurth:
> After we get some power measurements with capture.py script, we don't get the histogram files, right? I think we also need to add some other tests to cover the complete functionality of the tvla.py script (input may be hundreds of power traces and output may be the figures or their hashes).

Absolutely, this is just the first step in testing the full TVLA! Could you create an issue to collect aspects of TVLA that need testing, @abdullahvarici? We can then discuss and prioritize them together.

> Other than that, it worked in my system and looks good to me.

Great, thanks!

> In my PR, I have changed the commands in tvla_test.py to comply with the updated CLI of the tvla.py script: https://github.com/lowRISC/ot-sca/pull/92

Nice, let's focus on getting that merged as soon as this is approved and merged.

test/tvla_test.py
```diff
@@ -17,3 +19,24 @@ def test_help():
...
+def ttest_significant(ttest_trace) -> bool:
+    """Determine if a t-test trace contains a significant deviation from the mean."""
+    threshold = 4.5
+    abs_max = np.max(np.abs(ttest_trace), axis=3)
+    return np.any(abs_max > threshold)
+
+
+def test_general_leaking_histogram():
+    hist_path = TestDataPath('tvla_general/sha3_hist_leaking.npz')
+    tvla = TvlaCmd(Args(f'-g -i {hist_path}')).run()
```
vogelpi, line 33: Nit: add the `-m sha3` argument. It's not strictly required at the moment though.

vogelpi (approved):
Thanks, this looks good!

I would find it useful to add the command to run the test (`pytest test` in the main directory) into the commit message.

test/tvla_test.py
andreaskurth: Thanks :+1: changed

vogelpi: Excellent, thanks Andreas!

Issue #96, opened 2022-07-04
Automatic testing of tvla.py script


abdullahvarici:
With the PR https://github.com/lowRISC/ot-sca/pull/95, we can now load histograms and use them for general test. This is done automatically when a PR is created but it does not cover all functionality of the tvla.py script. We should also add an automatic test which gets power measurements, computes their t-values and compares with the expected values.

In addition there may be some other functionalities of the tvla.py script that may be tested. Maybe @vrozic, @vogelpi and @andreaskurth has some ideas?

andreaskurth:
Capturing the results of our [discussion yesterday](https://docs.google.com/document/d/1VZVsD8blMwPstgrOXidP6BeHS2Kiloda3lVryIlEHds/edit#heading=h.c3siwf58s61f):

**Highest priority** among the tests to be added for TVLA are:
- AES general and byte-based
- Stepping
- Filtering, not just starting from histograms

Furthermore, we should add tests that compare the **actual numerical values** (e.g., by comparing the t-test trace against expected data).

Test data should always be accompanied by a shell script containing the **commands used to capture/generate the data**. The primary purpose of that script is to make test data reproducible. This is required because some test data will have to be updated as we change the behavior of the capture and/or analysis code. If manual interaction with hardware peripherals is required to produce test data, the script does not have to be executable from start to beginning; rather, it should contain comments so that an engineer knows when and how to manually interact with hardware between running commands.

For some tests, it could be beneficial to generate **minimal, artificial traces** to be able to quickly pinpoint bugs. For example, one problem with real traces is that you cannot always know at which exact clock cycle leakage is expected. 

Finally, we should work with **compressed test data** to save LFS space and bandwidth. Note that `np.savez()` creates *uncompressed* `.npz` archives; use [`np.savez_compressed()`](https://numpy.org/doc/stable/reference/generated/numpy.savez_compressed.html#numpy.savez_compressed) to create `zipfile.ZIP_DEFLATED` `.npz` archives.

PR #97, opened 2022-07-06
Compress TVLA general test data

andreaskurth:
This commit compresses the existing test data for the general TVLA test
with `np.savez_compressed()`.  No changes are required to the code, which
can continue to load the data with a simple `np.load()`.  The total file
size of the existing tests is reduced from >150 MiB to <1 MiB.  The
execution time of the tests does not increase measurably.

Part of #96.


vogelpi (approved):
Excellent, thanks @andreaskurth ! This will become more important as we will add more tests.

PR #98, opened 2022-07-26
[otbn/ecdsa] Add capture support for ECDSA-256/384

bilgiday:
Initial support to capture side-channel traces from
OTBN ECDSA-256 and ECDSA-384 applications.

New capture commands `ecdsa-simple` and `ecdsa-stream`.

Supports only CW-Husky as the capture device.

* New files:
   * `cw/capture_ecdsa_cw310.yaml`: config file for ECDSA experiments

* Modified files:
   * `cw/capture.py`
      * Add new command `ecdsa-simple` to collect long ECDSA traces without using CW-Husky's streaming mode
      * Add new command `ecdsa-stream` to collect long ECDSA traces using CW-Husky's streaming mode
   * `cw/device.py`
      * Changes to handle PLL frequencies other than 100MHZ
      * Add `program_target()` for programming the firmware during ECDSA
experiments.

* Notes:
   * Tested on CW-310 with a CW-Husky

Signed-off-by: Bilgiday Yuce <bilgiday@opentitan.org>


vrozic:
Thanks @bilgiday for opening this PR and for making the changes.

Capturing works well when using binary `objs/ecc_serial_fpga_cw310.bin` 
I tested it using:
`./capture.py --cfg-file capture_ecdsa_cw310.yaml capture ecdsa-simple --num-traces 10`
and
`./capture.py --cfg-file capture_ecdsa_cw310.yaml capture ecdsa-stream --num-traces 10`
In both cases, the traces look correct.

However, when I try testing `objs/ecc384_serial_fpga_cw310.bin`, I get the timeout warnings
`Timeout in OpenADC capture(), no trigger seen!` and the capturing doesn't complete. Does this binary work in your setup?

cw/capture_ecdsa_cw310.yaml
```diff
@@ -0,0 +1,30 @@
+device:
+  fpga_bitstream: /mnt/chromeos/MyFiles/Downloads/lowrisc_systems_chip_earlgrey_cw310_0.1.repo5.bit
```
vrozic, line 2:
```suggestion
  fpga_bitstream: objs/lowrisc_systems_chip_earlgrey_cw310_0.1_ecdsa.bit
```

bilgiday: Thanks, fixed it now

bilgiday:
> Thanks @vrozic for testing them.

I think the problem is related to the following lines of `cw/capture_ecdsa_cw310.yaml`:
```
key_len_bytes: 32
plain_text_len_bytes: 32
output_len_bytes: 32
```
ECDSA-256 driver works with 32-B input/output while ECDSA-384 works with 48-B input/output.

To make things easier for the user we can use a separate yaml file for each application such as  `capture_ecdsa256_cw310.yaml` and `capture_ecdsa384_cw310.yaml`. Would it be okay?

vrozic:
> ECDSA-256 driver works with 32-B input/output while ECDSA-384 works with 48-B input/output.
> 
> To make things easier for the user we can use a separate yaml file for each application such as `capture_ecdsa256_cw310.yaml` and `capture_ecdsa384_cw310.yaml`. I can create these new files, would it be okay?

Ooops, of course. It is working well now. :)
Yes, having separate `.yaml` files is a good idea. It will help avoid any confusion in future.

vrozic (approved):
I've tested the setup and it all works OK.
I think that this is ready to merge.

PR #99, opened 2022-07-27
Add support for KMAC SW key masking

vrozic:
KMAC key shares are masked in SW using a 32-bit LFSR.
This commit enables user to specify the LFSR seed in the
configuration file. This feature is needed because unmasked
implementations don't support SW key masking. By re-seeding
LFSR to 0, SW key-masking can be effectively switched off.

Signed-off-by: Vladimir Rozic <vrozic@lowrisc.org>


vogelpi (approved):
LGTM, thanks @vrozic !

PR #100, opened 2022-09-27
[docker/podman] Rework container, switch to podman

vogelpi:
This PR contains two commits to update the Dockerfile and switch from docker to podman which doesn't require root permissions on the host. This is what we use on our workstation and we might do the same switch also for OpenTitan at some point. At the moment, the generated image is still compatible with docker.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,3 +1,4 @@
 *
 !python-requirements.txt
+!python-requirements-lint.txt
 !util/docker
--- a/doc/getting_started.md
+++ b/doc/getting_started.md
@@ -142,17 +142,20 @@ Alternatively, you can rebuild those binaries yourself from the
 
 #### Using the Docker Image
 
-**Note**: This is a WIP and currently supports only Linux hosts.
+**Notes**:
+- While the provided container image can be run using the Docker Engine, this
+  getting started guide relies on Podman instead to avoid requiring root
+  permissions on the host.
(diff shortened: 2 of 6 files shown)


andreaskurth (approved):
LGTM, thanks @vogelpi!

alphan:
LGTM but just wanted to point out that docker doesn't really _require_ root permissions on the host. I assume usb devices also work fine with `podman`, right?

vogelpi:
> LGTM but just wanted to point out that docker doesn't really _require_ root permissions on the host. I assume usb devices also work fine with `podman`, right?

Thanks for the feedback @alphan . I must have misunderstood the root permissions thing then, sorry. Yes, the USB devices also work with podman. Following your guidance, I did create some symlinks on our workstation based on the serial number. This allows having a container for the CW305 and one for the CW310, each with a separate Husky connected.

alphan:
> > LGTM but just wanted to point out that docker doesn't really _require_ root permissions on the host. I assume usb devices also work fine with `podman`, right?
> 
> Thanks for the feedback @alphan . I must have misunderstood the root permissions thing then, sorry. Yes, the USB devices also work with podman. Following your guidance, I did create some symlinks on our workstation based on the serial number. This allows having a container for the CW305 and one for the CW310, each with a separate Husky connected.

SG, thanks for confirming @vogelpi!

PR #101, opened 2022-09-27
[util/device.py] Wait for ADC lock instead of aborting after first trial

vogelpi:
The scope uses the clock provided from the target and depending on the target, it may take some more time for the ADC to actually lock on the clock. With this commit, we wait for up to 1.5s for the ADC to lock instead of just aborting the capture if the ADC hasn't locked yet when first checked.

### DIFF
--- a/cw/util/device.py
+++ b/cw/util/device.py
@@ -158,7 +158,16 @@ class OpenTitan(object):
         if not husky:
             scope.clock.reset_adc()
             time.sleep(0.5)
-        assert (scope.clock.adc_locked), "ADC failed to lock"
+
+        # Wait for ADC to lock.
+        ping_cnt = 0
+        while not scope.clock.adc_locked:
+            if ping_cnt == 3:
+                raise RuntimeError(
+                    f'ADC failed to lock (attempts: {ping_cnt}).')
+            ping_cnt += 1
+            time.sleep(0.5)
+
         return scope
 
     def initialize_target(self, programmer, firmware, baudrate, output_len,


vogelpi:
Hi @abdullahvarici ,
I think this is related to what you've discussed with @colinoflynn at CHES. I've just come across the same problem and it's solved by this PR.

cw/util/device.py
```diff
@@ -158,7 +158,16 @@ def initialize_scope(self, scope_gain, num_samples, offset, pll_frequency):
...
             scope.clock.reset_adc()
             time.sleep(0.5)
-        assert (scope.clock.adc_locked), "ADC failed to lock"
+
+        # Wait for ADC to lock.
+        ping_cnt = 0
+        while not scope.clock.adc_locked:
+            if ping_cnt == 3:
+                raise RuntimeError(
+                    f'"ADC failed to lock" (attempts: {ping_cnt}).')
```
andreaskurth, line 167: nit: double quotes seem unnecessary
```suggestion
                    f'ADC failed to lock (attempts: {ping_cnt}).')
```

andreaskurth (approved):
LGTM (with one minor nit), thanks!

cw/util/device.py
vogelpi: Done, thanks @andreaskurth !

colinoflynn:
@vogelpi glad you were able to replicated/solve this one! If it wasn't that I was running out of ideas...

vogelpi:
Hey @colinoflynn ,
thanks for your feedback! IIRC it was you who suggested exactly this solution :-)

abdullahvarici:
Hi @vogelpi. Thanks for solving it and thanks @colinoflynn for the idea :)

abdullahvarici:
I can still have the error `ADC failed to lock` sometimes when capturing traces. Powering off and on the board fixed the problem.

Issue #112, opened 2023-01-19
add doc for sha3 capture and tvla


m-temp:
@y-srini  mentioned in the ot-sca meeting that some doc regarding sha3 capture and tvla would be helpful.
I'll push a PR soon, but want to address https://github.com/lowRISC/ot-sca/issues/111 first.

y-srini:
@m-temp: 
In the SHA3 batch mode section, the capture parameter should be updated to "sha3-fvsr-data-batch"

./capture.py --cfg-file capture_sha3_cw310.yaml capture **sha3-fvsr-data-batch** --num-traces 100 --plot-traces 10

PR #114, opened 2023-01-19
[doc/sha3] add sha3 section to getting-started guide

m-temp:
This fixes https://github.com/lowRISC/ot-sca/issues/112

Would be good if s.o. could double check that the png is added correctly to lfs

### DIFF
--- a/doc/getting_started.md
+++ b/doc/getting_started.md
@@ -567,8 +567,8 @@ $ cd cw
 $ ./capture.py --cfg-file capture_kmac_cw310.yaml capture kmac-random --num-traces 100 --plot-traces 10
 ```
 
-The above command will send KMAC-128 requests with a fixed key and random 
-texts. In order to capture traces for DTR TVLA Section 5.3: "General Test: 
+The above command will send KMAC-128 requests with a fixed key and random
+texts. In order to capture traces for DTR TVLA Section 5.3: "General Test:
 Fixed-vs.-Random Key Datasets", following command may be used:
 ```console
 $ cd cw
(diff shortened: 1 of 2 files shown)


alphan (approved):
Thanks @m-temp. I didn't checkout your PR but github tells me that the png is stored with Git LFS when I view your commit:

Screenshot from 2023-01-19 11-36-55

y-srini:
@m-temp:
In the SHA3 batch mode section, the capture parameter should be updated to "sha3-fvsr-data-batch"

./capture.py --cfg-file capture_sha3_cw310.yaml capture sha3-fvsr-data-batch --num-traces 100 --plot-traces 10

m-temp:
@y-srini Thanks for having a second look. Should be fixed now.

doc/getting_started.md
```diff
@@ -581,6 +581,56 @@ browser.
...
+
+
+
+### SHA3 TVLA
+
+To perform the tvla just run
+```console
+$ cd cw
+$ ./tvla.py --cfg-file tvla_cfg_sha3.yaml run-tvla
+```
```
vrozic, line 625: This will run the general (non-specific) TVLA, which is the only one that we support right now.
It might be useful to mention that this analysis should be used only on Fixed-vs.-Random Datasets.

vrozic (approved):
Thanks @m-temp for the update.
I can confirm that the png is correctly added to the lfs.
Apart from my previous comment this looks ready to merge.

doc/getting_started.md
m-temp: Good point. Done.

PR #115, opened 2023-01-25
[OTBN] vertical attack initial work

m-temp:
This is a PR to bring together the works of Jade, Bilgiday and me

Co-authored-by: Jade Philipoom <jadep@google.com>
Signed-off-by: Michael Tempelmeier <michael.tempelmeier@gi-de.com>

Things to do:
- [x] Add TVLA support
    - [x] stimuli
    - [x] analysis script
- <del>[ ] Restructure code to support different code snippets</del>
    - <del>[ ] add select function to simple serial (python + c)</del>
    - <del>[ ] put "golden model" in a function to support different calculations</del>
- [x] Prepare and merge PR in opentitan repo ==> add binaries with specific hash id

### DIFF
--- a/cw/capture.py
+++ b/cw/capture.py
@@ -135,7 +135,8 @@ def capture_init(ctx, num_traces, plot_traces):
 
     # Key and plaintext generator
     ctx.obj.ktp = cw.ktp.Basic()
-    ctx.obj.ktp.key_len = cfg["capture"]["key_len_bytes"]
+    ctx.obj.ktp.fixed_key = False  # if true, then key never updates and setting...
+    ctx.obj.ktp.key_len = cfg["capture"]["key_len_bytes"]  # ...key_len has no effect
     ctx.obj.ktp.text_len = cfg["capture"]["plain_text_len_bytes"]
 
     ctx.obj.ot = initialize_capture(cfg["device"], cfg["capture"])
@@ -992,6 +993,173 @@ def kmac_fvsr_key_batch(ctx: typer.Context,
     capture_init(ctx, num_traces, plot_traces)
     capture_kmac_fvsr_key_batch(ctx.obj.ot, ctx.obj.ktp,
                                 ctx.obj.cfg["capture"], scope_type)
+    capture_end(ctx.obj.cfg)
+
+
+def capture_otbn_vertical(ot, ktp, fw_bin, pll_frequency, capture_cfg):
+    """Capture traces for ECDSA P256/P384 secret key generation.
+
+    Uses a fixed seed and generates several random masks. For the corresponding
+    driver, check:
+    <opentitan_repo_root>/sw/device/sca/ecc256_keygen_serial.c
+
+    Args:
+      ot: Initialized OpenTitan target.
+      ktp: Key and plaintext generator.
+      fw_bin: Firmware binary.
+      pll_frequency: Output frequency of the FPGA PLL.
+      capture_cfg: Capture configuration from the yaml file
+    """
+
+    # This determines if we want to reprogram the firmware
+    # before every sign operation.The default value is false.
+    reset_firmware = False
+
+    # OTBN operations are long. CW-Husky can store only 131070 samples
+    # in the non-stream mode.
+    fifo_size = 131070
+    if capture_cfg["num_samples"] > fifo_size:
+        raise RuntimeError('Current setup only supports up to 130k samples')
+
+    # Be sure we don't use the stream mode
+    if ot.scope._is_husky:
+        ot.scope.adc.stream_mode = False
+        ot.scope.adc.bits_per_sample = 12
+        ot.scope.adc.samples = fifo_size
+    else:
+        raise RuntimeError('Only CW-Husky is supported now')
+
+    # Create a cw project to keep the data and traces
+    project = cw.create_project(capture_cfg["project_name"], overwrite=True)
+
+    # Initialize some curve-dependent parameters.
+    if capture_cfg["curve"] == 'p256':
+        curve_order_n = 0xffffffff00000000ffffffffffffffffbce6faada7179e84f3b9cac2fc632551
+        # key_bytes = 256 // 8
+        seed_bytes = 320 // 8
+    else:
+        # TODO: add support for P384
+        raise NotImplementedError(f'Curve {capture_cfg["curve"]} is not supported')
+
+    # Check the lengths in the key/plaintext generator. In this case, "key"
+    # means seed and "plaintext" means mask.
+    if ktp.keyLen() != seed_bytes:
+        raise ValueError(f'Unexpected seed length: {ktp.keyLen()}.\n'
+                         f'Hint: set key len={seed_bytes} in the configuration file.')
+    if ktp.textLen() != seed_bytes:
+        raise ValueError(f'Unexpected mask length: {ktp.textLen()}.\n'
+                         f'Hint: set plaintext len={seed_bytes} in the configuration file.')
+
+    # Seed the RNG and generate a random fixed seed for all traces of the
+    # keygen operation.
+    seed = ktp.next_key()
+    print(f'seed = {seed.hex()}')
+    if len(seed) != seed_bytes:
+        raise ValueError(f'Seed length is {len(seed)}, expected {seed_bytes}')
+
+    # Expected key is `seed mod n`, where n is the order of the curve and
+    # `seed` is interpreted as little-endian.
+    expected_key = int.from_bytes(seed, byteorder='little') % curve_order_n
+
+    # Loop to collect each power trace
+    for _ in tqdm(range(capture_cfg["num_traces"]), desc='Capturing',
+                  ncols=80):
+
+        if reset_firmware:
+            ot.program_target(fw_bin, pll_frequency)
+
+        ot.scope.adc.offset = capture_cfg["offset"]
+
+        # Generate a new random mask for each trace.
+        mask = ktp.next_text()
+        tqdm.write(f'mask = {mask.hex()}')
+
+        # Set the seed.
+        ot.target.simpleserial_write('x', seed)
+
+        # Check for errors.
+        err = ot.target.read()
+        if err:
+            raise RuntimeError(f'Error writing seed: {err}')
+
+        # Arm the scope
+        ot.scope.arm()
+
+        # Send the mask and start the keygen operation.
+        ot.target.simpleserial_write('k', mask)
+
+        # Wait until operation is done.
+        ret = ot.scope.capture(poll_done=True)
+        # TODO: change this
+        # If getting inconsistent results (e.g. variable number of cycles),
+        # adding a sufficient sleep below here appears to fix things
+        # time.sleep(1)  # we cannot afford 1 second per run.
+        if ret:
+            raise RuntimeError('Timeout during capture')
+
+        # Check the number of cycles where the trigger signal was high.
+        cycles = ot.scope.adc.trig_count
+        tqdm.write("Observed number of cycles: %d" % cycles)
+
+        waves = ot.scope.get_last_trace(as_int=True)
+        # Read the output, unmask the key, and check if it matches
+        # expectations.
+        share0 = ot.target.simpleserial_read("r", seed_bytes, ack=False)
+        share1 = ot.target.simpleserial_read("r", seed_bytes, ack=False)
+        if share0 is None or share1 is None:
+            raise RuntimeError('Did not receive expected output.')
+
+        d0 = int.from_bytes(share0, byteorder='little')
+        d1 = int.from_bytes(share1, byteorder='little')
+        actual_key = (d0 + d1) % curve_order_n
+        if actual_key != expected_key:
+            raise RuntimeError('Bad generated key:\n'
+                               f'Expected: {hex(expected_key)}\n'
+                               f'Actual:   {hex(actual_key)}')
+
+        # Create a chipwhisperer trace object and save it to the project
+        # Args/fields of Trace object: waves, textin, textout, key
+        textout = share0 + share1
+        trace = Trace(waves, mask, textout, seed)
+        check_range(waves, ot.scope.adc.bits_per_sample)
+        project.traces.append(trace, dtype=np.uint16)
+
+    project.save()
+
+
+@app_capture.command()
+def otbn_vertical(ctx: typer.Context,
+                  num_traces: int = opt_num_traces,
+                  plot_traces: int = opt_plot_traces):
+    """Capture ECDSA secret key generation traces."""
+    capture_init(ctx, num_traces, plot_traces)
+
+    # OTBN's public-key operations might not fit into the sample buffer of the scope
+    # These two parameters allows users to conrol the sampling frequency
+    #
+    # `adc_mul` affects the clock frequency (clock_freq = adc_mul * pll_freq)
+    #
+    # `decimate` is the ADC downsampling factor that allows us to sample at
+    #  every `decimate` cycles.
+    if "adc_mul" in ctx.obj.cfg["capture"]:
+        ctx.obj.ot.scope.clock.adc_mul = ctx.obj.cfg["capture"]["adc_mul"]
+    if "decimate" in ctx.obj.cfg["capture"]:
+        ctx.obj.ot.scope.adc.decimate = ctx.obj.cfg["capture"]["decimate"]
+
+    # Print the params
+    print(
+        f'Target setup with clock frequency {ctx.obj.cfg["device"]["pll_frequency"]} MHz'
+    )
+    print(
+        f'Scope setup with sampling rate {ctx.obj.ot.scope.clock.adc_freq} S/s'
+    )
+
+    # Call the capture loop
+    capture_otbn_vertical(ctx.obj.ot,
+                          ctx.obj.ktp,
+                          ctx.obj.cfg["device"]["fw_bin"],
+                          ctx.obj.cfg["device"]["pll_frequency"],
+                          ctx.obj.cfg["capture"])
     capture_end(ctx.obj.cfg)
 
 
--- a/cw/capture.py
+++ b/cw/capture.py
@@ -135,8 +135,11 @@ def capture_init(ctx, num_traces, plot_traces):
 
     # Key and plaintext generator
     ctx.obj.ktp = cw.ktp.Basic()
-    ctx.obj.ktp.fixed_key = False  # if true, then key never updates and setting...
-    ctx.obj.ktp.key_len = cfg["capture"]["key_len_bytes"]  # ...key_len has no effect
+    # This is a workaroung for https://github.com/lowRISC/ot-sca/issues/116
+    if "use_fixed_key_iter" in cfg["capture"]:  # for backwarts compatibility
+        ctx.obj.ktp.fixed_key = cfg["capture"]["use_fixed_key_iter"]
+    # ktp.key_len is only evaluated if ktp.fixed_key is set to False
+    ctx.obj.ktp.key_len = cfg["capture"]["key_len_bytes"]
     ctx.obj.ktp.text_len = cfg["capture"]["plain_text_len_bytes"]
 
     ctx.obj.ot = initialize_capture(cfg["device"], cfg["capture"])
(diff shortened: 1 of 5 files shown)


cw/capture.py
```diff
@@ -135,6 +135,7 @@ def capture_init(ctx, num_traces, plot_traces):
 
     # Key and plaintext generator
     ctx.obj.ktp = cw.ktp.Basic()
+    ctx.obj.ktp.fixed_key = False # if true, then key never updates and setting key_len has no effect
```
m-temp, line 138: This is a small change, but might have a huge impact!
By default the ktp.next always outputs the same key. I saw multiple workarounds throughout this project like using the kpt.plaintext_next as a random number generator for both key and plaintext, but I'm not sure if there are parts that amuse default behavior....

We should discuss this change.

cw/capture.py
```diff
@@ -991,6 +992,182 @@ def kmac_fvsr_key_batch(ctx: typer.Context,
     capture_end(ctx.obj.cfg)
 
 
+def capture_ecdsa_secret_keygen(ot, ktp, fw_bin, pll_frequency, capture_cfg):
```
vogelpi, line 995: I suggest to rename this to something more generic, e.g. `capture_otbn_vertical()`. The reason for this is that `keygen` is just one/the first code snippet we agreed to do a vertical analysis for. In the future, we most likely will analyze also other snippets for which the settings will be very similar (but pretty different from analyzing full ECDSA). In that sense vertical analysis of OTBN code snippets is far closer to analyzing AES/KMAC than ECDSA. We could even do batch captures here but shouldn't need streaming. Capture rate should be 200 MHz.

 What might change for other snippets and we should try to keep things flexible in this regard:
- number of samples per trace (but they should be far below the max 132k samples)
- offsets
- number of inputs and length of inputs
- output length

cw/capture.py
```diff
@@ -991,6 +992,182 @@ def kmac_fvsr_key_batch(ctx: typer.Context,
...
+
+    # OTBN operations are long. CW-Husky can store only 131070 samples
+    # in the non-stream mode. In case we want to collect more samples,
+    # we can run the operation multiple times with different adc_offset
+    # values.
+    # The trace from each run is kept in a section, and all sections are
+    # concatenated to create the final trace.
+    fifo_size = 131070
+    num_sections = capture_cfg["num_samples"] // fifo_size
+    print(f"num_sections = {num_sections}")
```
vogelpi, line 1021: I think this shouldn't be needed. Maybe it's best to start from the capture routine used for SHA3/KMAC for this. The code snippets should be small. If if they are 500 instructions and we oversample by 20x, the individual trace will just be 10000 samples.

cw/util/device.py
```diff
@@ -76,8 +76,9 @@ def initialize_fpga(self, fpga, bitstream, pll_frequency):
         # bitstream.
         # Unfortunately, this doesn't seem to work for the CW310 yet,
         # see https://github.com/lowRISC/ot-sca/issues/48.
+        # This should work now. Issue is closed
```
vogelpi, line 79: I think if the issues is resolved we should actually remove this reference here. :-)

cw/util/device.py
```diff
@@ -76,8 +76,9 @@ def initialize_fpga(self, fpga, bitstream, pll_frequency):
...
         # TODO: We should have this in the CLI.
-        force_programming = False
+        force_programming = True
```
vogelpi, line 81: IIRC this will cause the FPGA to always be re-programmed which we actually don't want no?

vogelpi (commented):
Hi @m-temp ,
thanks for the PR!
I had a first brief look now. I think we should try to keep this capture setup as generic as possible because we're going to use it for other code snippets than keygen as well. Also, since the captures are short we don't need some of the infrastructure used in for ECDSA captures. The whole capture and analysis is much more similar to KMAC/AES. Maybe it's best to discuss this tomorrow, I've set up a little sync meeting.

cw/capture.py
m-temp: This will be resolved in https://github.com/lowRISC/ot-sca/issues/116
For this PR we will use a workaround with backwards compatibility.

m-temp:
I've added a first fixed-vs-random tvla version. There still a couple of todos like magic numbers etc.
Major points are:
 - finding good parameter sets (out of memory vs enough data)
 - add a batch version
 - restructure the code

m-temp:
To expedite cooperation, I mark this PR as ready to merge

It adds:
  - basic fvsr support for otbn vertical
  - ctrl+c to store capture results
  - config outputs to tvla plots

A brief test confirmed that sha3, kmac, aes are not broken.

What is still missing:
  - doc for otbn vertical
  - a clear commit id for the binaries. (we will add this with a separate PR that also improves the software running on the device.)

cw/capture.py
```diff
@@ -991,6 +1008,235 @@ def kmac_fvsr_key_batch(ctx: typer.Context,
...
+    seed_fixed = ktp.next_key()
+    print(f'fixed seed = {seed_fixed.hex()}')
+    if len(seed_fixed) != seed_bytes:
+        raise ValueError(f'Fixed seed length is {len(seed_fixed)}, expected {seed_bytes}')
+
+    # Expected key is `seed mod n`, where n is the order of the curve and
+    # `seed` is interpreted as little-endian.
+    expected_fixed_key = int.from_bytes(seed_fixed, byteorder='little') % curve_order_n
+
+    # register ctrl-c handler to not loose already recorded traces if measurement is aborted
```
abdullahvarici, line 1085: Typo
```suggestion
    # register ctrl-c handler to not lose already recorded traces if measurement is aborted
```

cw/capture.py
```diff
@@ -991,6 +1008,235 @@ def kmac_fvsr_key_batch(ctx: typer.Context,
...
+
+            # Send the mask and start the keygen operation.
+            ot.target.simpleserial_write('k', mask)
+
+            # Wait until operation is done.
+            ret = ot.scope.capture(poll_done=True)
+            if ret:
+                raise RuntimeError('Timeout during capture')
+
+            # Check the number of cycles where the trigger signal was high.
```
abdullahvarici, line 1128: Should we compare it with an expected number instead of printing?

cw/capture.py
```diff
@@ -991,6 +1008,235 @@ def kmac_fvsr_key_batch(ctx: typer.Context,
...
+        if capture_cfg["masks_off"] is True:
+            # Use a constant mask for each trace
+            mask = bytearray(capture_cfg["plain_text_len_bytes"])  # all zeros
+        else:
+            # Generate a new random mask for each trace.
+            mask = ktp.next_text()
+        tqdm.write("Starting new trace....")
+        tqdm.write(f'mask   = {mask.hex()}')
+
+        if sample_fixed:
```
abdullahvarici, line 1107: The same code is shared in if and else part. Shouldn't we put it after the if else statement?

cw/capture.py
```diff
@@ -991,6 +1008,235 @@ def kmac_fvsr_key_batch(ctx: typer.Context,
...
+
+        sample_fixed = random.randint(0, 1)
+        if actual_key != expected_key:
+            raise RuntimeError('Bad generated key:\n'
+                               f'Expected: {hex(expected_key)}\n'
+                               f'Actual:   {hex(actual_key)}')
+
+        # Create a chipwhisperer trace object and save it to the project
+        # Args/fields of Trace object: waves, textin, textout, key
+        textout = share0 + share1  # concardinate bytearrays
```
abdullahvarici, line 1196: Typo
```suggestion
        textout = share0 + share1  # concatenate bytearrays
```

cw/capture.py
```diff
@@ -991,6 +1008,235 @@ def kmac_fvsr_key_batch(ctx: typer.Context,
     capture_end(ctx.obj.cfg)
 
 
+def capture_otbn_vertical(ot, ktp, fw_bin, pll_frequency, capture_cfg):
```
abdullahvarici, line 1011: I think we should either make the function name specific or move algorithm specific parts (generating input and comparing result) to specific functions.

cw/capture.py
```diff
@@ -991,6 +1008,235 @@ def kmac_fvsr_key_batch(ctx: typer.Context,
...
+@app_capture.command()
+def otbn_vertical(ctx: typer.Context,
+                  num_traces: int = opt_num_traces,
+                  plot_traces: int = opt_plot_traces):
+    """Capture ECDSA secret key generation traces."""
+    capture_init(ctx, num_traces, plot_traces)
+
+    # OTBN's public-key operations might not fit into the sample buffer of the scope
+    # These two parameters allows users to conrol the sampling frequency
+    #
```
abdullahvarici, line 1213: Do we need this empty comment line?

cw/capture.py
```diff
@@ -991,6 +1008,235 @@ def kmac_fvsr_key_batch(ctx: typer.Context,
...
+    # `adc_mul` affects the clock frequency (clock_freq = adc_mul * pll_freq)
+    #
```
abdullahvarici, line 1215: Do we need this empty comment line?

cw/tvla.py
```diff
@@ -336,15 +336,18 @@ def run_tvla(ctx: typer.Context):
                     level=log.INFO,
                     force=True,)
 
-    if cfg["mode"] != "kmac" and cfg["mode"] != "aes" and cfg["mode"] != "sha3":
+    if (cfg["mode"] != "kmac" and cfg["mode"] != "aes" and cfg["mode"] != "sha3" and
+            cfg["mode"] != "sha3" and cfg["mode"] != "otbn"):
```
abdullahvarici, line 340: Duplicate "sha3" check!

abdullahvarici (commented):
Hello! Thanks for the PR. Sorry, I don't have any hardware to try it out for now. I think, in general it looks good. I've added a couple of feedbacks.

cw/capture.py
m-temp: This is a good point. I totally agree.
It was name specific at first, but then we discussed in one of the sca meetings that it would be good to eventually support multiple algorithms. While moving algorithmic specific parts to specific functions, I came to the conclusion that this should be done at a later point when we actually have multiple algorithms.

So what's the best way for now?
- Change the name (here and in the config files) back, merge the PR, and then change it again, when we actually have multiple algorithms?
- Leave it as it is, knowing that there will be multiple algorithms soon? 
- Move the algorithmic specific parts to specific functions? (actually, I don't want to do this in this PR, as the PR is already too large)

@vogelpi what do you think?

cw/capture.py
```diff
@@ -60,6 +63,16 @@ def create_cw_segmented(ot, capture_cfg):
 }
 
 
+# handler for ctrl-c keyboard interrupts
+# saves capture project before exiting in case abort is intended
+# needs to be registered in every capture function before capture loop
+def abort_handler(project, sig, frame):
```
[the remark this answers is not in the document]
wettermo: Thanks, you're right this looks more consistent. I changed it.

m-temp:
Thanks @abdullahvarici for your review and sry for all those typos in the code. I will address them.

cw/capture.py
```diff
@@ -991,6 +1008,235 @@ def kmac_fvsr_key_batch(ctx: typer.Context,
...
+    # For backwards compatibility this must be set in the capture config file.
+    # This is a workaroung for https://github.com/lowRISC/ot-sca/issues/116
+    if "use_fixed_key_iter" not in capture_cfg:
+        raise RuntimeError('use_fixed_key_iter not se set!')
+    if capture_cfg["use_fixed_key_iter"] is not False:
+        raise RuntimeError('use_fixed_key_iter must be set to false!')
+
+    # Currently, we need to reprogram the firmware before every sign operation
+    # TODO: Investigate if this is needed.
+    reset_firmware = False
```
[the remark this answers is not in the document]
m-temp: Good point. I think @bilgiday investigated it and it is not needed anymore. Should we keep the option?
I'll change the comment, but removing the option should be done in a different PR

cw/capture.py
```diff
@@ -60,6 +63,16 @@ def create_cw_segmented(ot, capture_cfg):
...
 
 
+def abort_handler(project, sig, frame):
+    """ Handler for ctrl-c keyboard interrupts:
+        Saves capture project before exiting, in case abort is intended.
+        Needs to be registered in every capture function before capture loop. """
+    if project is not None:
+        print("caught keyboard interrupt -> saving project (traces)...")
+        project.save()
+    sys.exit(0)
```
vogelpi, line 73: This has to go into a separate PR, because it's unrelated to OTBN vertical analysis. Instead it's useful for any captures and a great addition actually, thanks for adding this!

cw/capture.py
vogelpi: Let's stick to the generic name and move things around as we add support for other snippets. I also remember having discussed this before as @m-temp pointed out.

cw/capture.py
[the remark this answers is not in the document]
vogelpi: Your proposal sounds good @m-temp.

cw/tvla_cfg_otbn.yaml
```diff
@@ -0,0 +1,18 @@
...
+input_file: null
+output_file: null
+number_of_steps: 1
+ttest_step_file: null
+plot_figures: true
+general_test: true
+mode: otbn
+key_len_bytes: 40
+sample_start: null
+num_samples : 4000
```
vogelpi, line 18: Missing newline at EOF

cw/tvla.py
```diff
@@ -504,9 +535,12 @@ def run_tvla(ctx: typer.Context):
...
                 if project.waves[0].dtype == 'uint16':
                     traces = np.empty((num_traces, num_samples), dtype=np.uint16)
+                    log.info(f"Will use samples from {sample_start} to {sample_start+num_samples}")
                     for i_trace in range(num_traces):
-                        traces[i_trace] = project.waves[i_trace + trace_start]
-                else:
+                        traces[i_trace] = project.waves[i_trace +
+                                                        trace_start][sample_start:sample_start +
+                                                                     num_samples]
+                else:  # FIXME: The wave slicing is only tested for otbn
```
vogelpi, line 543: This is a very nice feature. Thanks for adding this!

It will also be very useful for other targets. It's not expected to test this for all different targets (there are simply too many options and we have other priorities). However, I would appreciate if you changed the code such that there is simply an assertion to check that when sample_start and sample_end are set, also otbn is selected and provide a comment where this check is done. This way we have to change much less code later on when enabling it e.g. for AES or KMAC.

cw/tvla.py
```diff
@@ -883,12 +883,16 @@ def run_tvla(ctx: typer.Context):
...
+                "TVLA of " + cfg["project_file"] + '\n' + "No. of traces: " + str(num_traces))
+            axs[0].plot(xaxs, single_trace, "k")
+            axs[0].set_ylabel("trace")
             for i_order in range(num_orders):
-                axs[1 + i_order].plot(ttest_trace[i_order, 0, 0], "k")
-                axs[1 + i_order].plot(c * threshold, "r")
-                axs[1 + i_order].plot(-threshold * c, "r")
+                axs[1 + i_order].plot(xaxs, ttest_trace[i_order, 0, 0], "k")
+                axs[1 + i_order].plot(xaxs, c * threshold, "r")
+                axs[1 + i_order].plot(xaxs, -threshold * c, "r")
```
vogelpi, line 895: This commit needs to be merged with the one adding support for sample_start and num_samples. Also these two things should go into a separate PR.

cw/capture.py
m-temp: Not necessarily, but it's the same structure as in `ecdsa_simple` and `ecdsa_stream`.
Should I remove it there as well? Or better do that in a different PR?

cw/capture.py
m-temp: see above.

cw/tvla.py
```diff
@@ -883,12 +883,16 @@ def run_tvla(ctx: typer.Context):
         else:
             #
             c = np.ones(num_samples)
+            xaxs = range(sample_start, sample_start + num_samples)
             fig, axs = plt.subplots(3, sharex=True)
-            axs[0].plot(single_trace, "k")
+            axs[0].set_title(
+                "TVLA of " + cfg["project_file"] + '\n' + "No. of traces: " + str(num_traces))
```
vogelpi, line 889: I think just this line here actually belongs into this very commit.

vogelpi (commented):
Overall this looks good and we should move forward. Many thanks for all the work you've put into this @m-temp and @wettermo .

In addition to the OTBN vertical support, this PR contains a few very nice additions that should be factored out into separate PRs. These additions are uncontroversial and are unrelated to OTBN vertical support.

vogelpi:
Also updating the bitstreams and existing binaries can be factored out. Remember the goal is always to have as small PRs and commit as possible. This speeds up the review and allows us to progress faster as a team.

cw/capture.py
```diff
@@ -135,7 +148,11 @@ def capture_init(ctx, num_traces, plot_traces):
 
     # Key and plaintext generator
     ctx.obj.ktp = cw.ktp.Basic()
-    ctx.obj.ktp.key_len = cfg["capture"]["key_len_bytes"]
+    # This is a workaroung for https://github.com/lowRISC/ot-sca/issues/116
+    if "use_fixed_key_iter" in cfg["capture"]:  # for backwarts compatibility
+        ctx.obj.ktp.fixed_key = cfg["capture"]["use_fixed_key_iter"]
+        # if fixed_key = true, then key never updates and setting...
+    ctx.obj.ktp.key_len = cfg["capture"]["key_len_bytes"]  # ...key_len has no effect
```
vrozic, line 155: nit: This is a bit strange way of writing comments. I would avoid having code and comment in the same line.

cw/capture.py
vrozic: I would leave it as is for now and do the name change in a fallow-up PR.
We are likely to have multiple capture functions related to otbn vertical attacks, so maybe we should discuss and agree on a naming strategy.

cw/capture.py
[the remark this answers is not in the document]
bilgiday: With the new versions of the device-side firmware, we don't need it. I just kept it here as an option just in case. 

After we merge this PR, I can create a new one to clean `reset_firmware` options from the capture.py. Does that sound good?

cw/capture.py
bilgiday: I was keeping them for the readability. If it violates the styling guidelines, we can remove them with the `reset_firmware` PR proposed in [this comment](https://github.com/lowRISC/ot-sca/pull/115#discussion_r1118735037) above.

cw/capture.py
bilgiday: Thanks @m-temp and @wettermo!

m-temp: I will remove that from the current PR

cw/capture.py
m-temp: We could do that, however the actual number of cycles varies by `+-2`  cycles.
If needed, let's do that in another PR.

cw/tvla.py
m-temp: wilco

cw/tvla.py
m-temp: wilco

cw/tvla.py
m-temp: Hmm let's put it also in the other PR, where we change the plotting

m-temp:
Thx all for the reviews. I changed it accordingly and will file a couple of PRs for the remaining features.

cw/capture.py
```diff
@@ -991,6 +1008,235 @@ def kmac_fvsr_key_batch(ctx: typer.Context,
...
+    if capture_cfg["num_samples"] > fifo_size:
+        raise RuntimeError('Current setup only supports up to 130k samples')
+
+    # Be sure we don't use the stream mode
+    if ot.scope._is_husky:
+        ot.scope.adc.stream_mode = False
+        ot.scope.adc.bits_per_sample = 12
+        ot.scope.adc.samples = capture_cfg["num_samples"]
+    else:
+        # TODO: Add cw-lite support
```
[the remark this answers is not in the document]
vogelpi: Hm, I am not sure I understand your comment. I think we don't want to have a TODO to Add CW-Lite support, because we're never going to add it but to error out if CW-Lite is detected is the best thing to do here.

vogelpi (approved):
Thanks for addressing the comments. This looks good!

Two minor nits:
1. There is still a TODO for the CW-Lite, but we won't add support for this. Having the script to error out for something else than Husky still desirable though.
2. The PR also adds a new bitstream in the first commit. Is this really needed? We wanted to add the new bistream with the different URND behavior in a separate PR I believe.

m-temp:
>     1. There is still a TODO for the CW-Lite, but we won't add support for this. Having the script to error out for something else than Husky still desirable though.
Sry. I missed that. This was a leftover from Jade. I'll remove the comment, but keep the if-else-branch 

>     2. The PR also adds a new bitstream in the first commit. Is this really needed? We wanted to add the new bistream with the different URND behavior in a separate PR I believe.
Ah okay, I thought this was only related to the second bins, but you are right, I delete the old bins and file a separate PR for the new ones.

PR #120, opened 2023-02-28
[tvla] Various small fixes

vogelpi:
This PR contains three small fixes for TVLA. For details refer to the commit messages.

### DIFF
--- a/cw/tvla.py
+++ b/cw/tvla.py
@@ -970,23 +970,31 @@ def run_tvla(ctx: typer.Context):
                     # The widnow width is 100 samples + 40 samples extended on each side.
                     half_window = samples_per_rnd // 2 + 40
 
-                    samples = range(max(rnd_offset + (rnd_list[i_rnd] * samples_per_rnd) -
-                                        half_window, 0),
-                                    min(rnd_offset + (rnd_list[i_rnd] * samples_per_rnd) +
-                                        half_window, num_samples))
-
+                    samples = {"aes": range(max(rnd_offset + (rnd_list[i_rnd] * samples_per_rnd) -
+                                            half_window, 0),
+                                            min(rnd_offset + (rnd_list[i_rnd] * samples_per_rnd) +
+                                                half_window, num_samples))}
                 else:
-                    if cfg["mode"] == "aes":
-                        # Simply plot everything.
-                        samples = range(0, num_samples)
-                    else:
-                        # For now, let's focus on the key absorption only.
-                        samples = range(520, 2460)
-                        # Numbers for the eval branch:
+                    # Even if cfg["sample_start"] is specified and cfg["num_samples"] is not,
+                    # num_samples is set to a valid value and the code below has valid inputs.
+                    range_specified = True if \
+                        (("sample_start" in cfg and cfg["sample_start"] is not None) or
+                         ("num_samples" in cfg and cfg["num_samples"] is not None)) else False
+
+                    samples = {
+                        # Simply plot all samples within the selected range.
+                        "aes": range(0, num_samples),
+                        # Plot samples within key absorption phase, unless a range is specified.
+                        "kmac": range(0, num_samples) if range_specified else range(520, 2460),
+                        # Plot samples within actual SHA3 phase, unless a range is specified.
+                        "sha3": range(0, num_samples) if range_specified else range(1150, 3150),
+                        # Simply plot all samples within the selected range.
+                        "otbn": range(0, num_samples)
+                    }
 
                 for i_order in range(num_orders):
                     for i_byte in range(num_bytes):
-                        for i_sample in samples:
+                        for i_sample in samples[cfg["mode"]]:
                             axs[i_order].plot(ttest_step[i_order,
                                                          rnd_ext[i_rnd],
                                                          byte_ext[i_byte],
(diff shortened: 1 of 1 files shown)


m-temp (commented):
I think this looks good, but I would add some comments to make it more obvious.

cw/tvla.py
```diff
@@ -904,10 +904,11 @@ def run_tvla(ctx: typer.Context):
                     if cfg["mode"] == "aes":
                         # Simply plot everything.
                         samples = range(0, num_samples)
-                    else:
+                    elif cfg["mode"] == "kmac":
                         # For now, let's focus on the key absorption only.
                         samples = range(520, 2460)
-                        # Numbers for the eval branch:
+                    else:
+                        samples = range(100, 3200)
```
m-temp, line 912: I think we should add a comment to the file explaining the numbers.
It might seem clear now, but if later someone else has a look at these numbers a comment would help to understand the code faster. Especially for the else-case.

cw/tvla.py
```diff
@@ -856,11 +856,15 @@ def run_tvla(ctx: typer.Context):
 
             log.info("Plotting T-test Statistics vs. Number of Traces, this may take a while.")
             # Determine resolution.
-            xres_vec = [10e6, 1e6, 100e3, 10e3, 1e3]
+            xres_vec = [10e6, 1e6, 100e3, 10e3, 1e3, 100, 1]
```
m-temp, line 859: nit: out of curiosity why do we skip 1e2? I don't think this is really an issue, but on the first glance it looks like there is something missing ;)

cw/tvla.py
```diff
@@ -865,11 +856,15 @@ def run_tvla(ctx: typer.Context):
...
             for xres in xres_vec:
                 if int(np.around(trace_end_vec[0] / xres)) > 0:
```
vrozic, line 861: It seems that we have a pre-existing problem in the `if` condition. It would be appropriate to fix it with this commit.

`trace_end_vec[0] ` should be changed to `(trace_end_vec[0] - trace_start_vec[0])`

E.g. Currently if I run TVLA on traces 4000-5000 with `number_of_steps: 10`, the x-axis is indexed in multiples of `1k ` when it should be in multiples of `100`.

cw/tvla.py
```diff
@@ -865,11 +856,15 @@ def run_tvla(ctx: typer.Context):
...
-                    xres_label = str(int(xres / 1e6)) + 'M' if xres >= 1e6 else \
-                        str(int(xres / 1e3)) + 'k'
+                    if xres >= 1e6:
```
vrozic, line 862: It seems that we have a pre-existing problem in the `if` condition in line 861. It would be appropriate to fix it with this commit.

`trace_end_vec[0] ` should be changed to `(trace_end_vec[0] - trace_start_vec[0])`

E.g. Currently if I run TVLA on traces 4000-5000 with `number_of_steps: 10`, the x-axis is indexed in multiples of `1k ` when it should be in multiples of `100`.

cw/tvla.py
vrozic: I think it's because the x-axis still looks nice when we have two-digit numbers. It's only with three-digit numbers that it begins to look cluttered.
Maybe we should remove `10e3` and `10e6` to be consistent.

cw/tvla.py
vrozic: This piece of code has a potential to get more confusing as we add more modes.
Perhaps it's time to simplify it with a dictionary indexed by `cfg["mode"]`?

vogelpi: Good points, I'll add a comment and try to do a dictionary.

cw/tvla.py
vogelpi: It's because I find the `M` and `k` suffices are very easy to interpret. I am always struggling with the 1e1, 10e1, 1e2 notation etc.

Also two and 3-digit labels are still kind of okayish to read. Let me add the `10` as well.

cw/tvla.py
vogelpi: Uh, very good catch @vrozic ! Let me change that.

cw/tvla.py
vogelpi: Done.

cw/tvla.py
vogelpi: Done.

cw/tvla.py
vogelpi: Done.

vogelpi:
I've rebased the PR and tried to address your feedback @m-temp and @vrozic , PTAL.

cw/tvla.py
```diff
@@ -970,23 +970,29 @@ def run_tvla(ctx: typer.Context):
...
+
+                    samples = {
+                        # Simply plot all samples within the selected range.
+                        "aes": range(0, num_samples),
+                        # Plot samples within key absorption phase, unless a range is specified.
+                        "kmac": range(0, num_samples) if range_specified else range(520, 2460),
+                        # Plot samples within actual SHA3 phase, unless a range is specified.
+                        "sha3": range(0, num_samples) if range_specified else range(1150, 3150),
+                        # Simply plot all samples within the selected range.
+                        "otbn": range(0, num_samples)
```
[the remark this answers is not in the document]
wettermo: Yes, but these lines of code shouldn't be affected by my changes. So don't mind me. My changes anyway are still kind of a work in progress. I'll probably open a PR for them next week.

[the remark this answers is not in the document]
vogelpi: Well, if you specify `sample_start` > 0 and/or `num_samples` < `len(waves[0])` the whole TVLA is computed on this range only. So `ttest_step` will only contain samples in that range. `ttest_step[:,:,:...,0]` corresponds to the t-test result at `sample_start` in the original / un-sliced trace. `range(sample_start, sample_start+num_samples)` would give you the wrong samples and might even be out of range of `ttest_step`.

cw/tvla.py
```diff
@@ -970,23 +970,29 @@ def run_tvla(ctx: typer.Context):
...
-                    if cfg["mode"] == "aes":
-                        # Simply plot everything.
-                        samples = range(0, num_samples)
-                    else:
-                        # For now, let's focus on the key absorption only.
-                        samples = range(520, 2460)
-                        # Numbers for the eval branch:
+                    range_specified = True if \
+                        (("sample_start" in cfg and cfg["sample_start"] is not None) or
+                         ("num_samples" in cfg and cfg["num_samples"] is not None)) else False
```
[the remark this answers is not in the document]
vogelpi: SG

cw/tvla.py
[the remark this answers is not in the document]
m-temp: ah I see. Having a look at this line only, the code is not very intuitive. Should we ever have time for a code-cleanup this could be part of it ;)

m-temp (approved):
LGTM

PR #121, opened 2023-02-28
[tvla] Add function for plotting average and variance traces and diffs

vogelpi:
These functions are particularly useful when debugging e.g. cases where the t-test results are not centered around 0.

Below figures show some example plots for SHA3 where I used this to debug a "fake" leakage issue.

1. The average fixed and random trace. They cannot really be distinguished here.
bokeh_plot

2. The difference between the fixed and random trace average. We see a big peak at the very beginning where the plaintext is loaded. Then the difference slowly and almost linearly decreases while the target is completely idle. This is when the power rails recharge from the initial plaintext loading.
bokeh_plot(1)

3. The variance for the fixed and random trace set.
bokeh_plot(2)

4. The difference between the fixed and random trace variance.
bokeh_plot(3)

5. The resulting TVLA plot. We can see the resulting fake leakage during the idle period and spanning into the SHA3 processing where it covers potential real leakage. These newly added figures are useful to debug such issues.
tvla

FYI: to reduce the fake leakage one has to change the fixed plaintext to something more uniform, for details see #122 .

### DIFF
--- a/cw/tvla.py
+++ b/cw/tvla.py
@@ -19,6 +19,8 @@ from chipwhisperer.analyzer import aes_funcs
 from joblib import Parallel, delayed
 from scipy.stats import ttest_ind_from_stats
 
+from util import plot
+
 app = typer.Typer(add_completion=False)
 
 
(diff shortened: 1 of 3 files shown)


cw/tvla.py
```diff
@@ -48,6 +50,38 @@ def bit_count(int_no):
     return c
 
 
+def plot_fixed_vs_random(traces, leakage):
```
vrozic, line 53: I suggest to change this name into something more precise, e.g. `plot_stats_fvsr` or `plot_statistics_fvsr`.

vogelpi: Sounds good. Let me do this.

cw/tvla.py
```diff
@@ -648,6 +682,10 @@ def run_tvla(ctx: typer.Context):
                 for i in range(num_traces):
                     leakage[i] = np.array_equal(fixed_key, keys[i])
 
+            # Uncomment the function call below for debugging e.g. when the t-test results aren't
+            # centered around 0.
+            # plot_fixed_vs_random(traces, leakage)
```
vrozic, line 687: Eventually we will probably want to add the plot_statistics argument and to modify tvla_cfg.yaml files.
Perhaps add a `#TO DO:` comment here so we don't forget

cw/tvla.py
vogelpi: Done.

cw/tvla.py
vogelpi: I think we really just need that for debugging purposes and having yet another argument for this is probably not worth it.

Also, if we really wanted that, we should also switch the function implementation to compute the means and variances from the histogram data. Right now it may consume quite a lot of memory and compute. It's really just for debugging. Thus I would say let's not worry about that.

vrozic (approved):
Thanks @vogelpi ,
this is a very useful function.

vogelpi:
Thanks for your review @vrozic , I am merging this.

PR #122, opened 2023-02-28
[capture] Fixes for SHA3 capture

vogelpi:
This PR contains two commits for the SHA3 capture. The second one is very relevant, I think this might potentially resolve also the issue @y-srini has been facing. I've found this out using the plots added with PR #121 .

### DIFF
--- a/cw/capture.py
+++ b/cw/capture.py
@@ -624,8 +624,28 @@ def capture_sha3_fvsr_data_batch(ot, ktp, capture_cfg, scope_type):
     if ack_ret is None:
         raise Exception("Batch mode acknowledge error: Device and host not in sync")
 
-    plaintext_fixed = bytearray([0x81, 0x1E, 0x37, 0x31, 0xB0, 0x12, 0x0A, 0x78,
-                                 0x42, 0x78, 0x1E, 0x22, 0xB2, 0x5C, 0xDD, 0xF9])
+    # Value defined under Section 5.3 in the derived test requirements (DTR) for TVLA.
+    plaintext_fixed = bytearray([0xDA, 0x39, 0xA3, 0xEE, 0x5E, 0x6B, 0x4B, 0x0D,
+                                 0x32, 0x55, 0xBF, 0xEF, 0x95, 0x60, 0x18, 0x90])
+
+    # Note that - at least on FPGA - the DTR value above may lead to "fake" leakage as for the
+    # fixed trace set, the number of bits set in the first (37) and second 64-bit word (31), as
+    # well as in the Hamming distance between the two (30) is different from the statistical
+    # mean (32). As a result, the loading of the fixed message into the SHA3 core on average
+    # discharges the power rails slightly less than loading a random message. Until the SHA3 core
+    # starts processing, the power rails will recharge but they might not be able to reach the same
+    # levels for the fixed and random trace set, potentially leading to a small vertical offset
+    # between the two trace sets. This offset is detectable by TVLA and covers actual leakage
+    # happening during the SHA3 processing. The effect is most easliy visible between loading the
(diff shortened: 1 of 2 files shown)


vogelpi:
FYI: I am not yet entirely sure why only observed this fake leakage with new bitstreams and not the one in the repo currently. I created multiple new bitstreams and they all showed the same behavior but to a different extent. It could be due to P&R variation. I observed that over the last couple of months LUT utilization increased again to almost 85% with masked KMAC . But there were no RTL changes in KMAC and relevant prims which could have caused this.

Also with the new fixed plaintext the behavior occurs but it's much weakger and becomes visible at higher number of traces.

cw/capture.py
```diff
@@ -593,8 +593,27 @@ def capture_sha3_fvsr_data_batch(ot, ktp, capture_cfg, scope_type):
...
+    # levels for the fixed and random trace set, potentially leading to a small vertical offset
+    # between the two trace sets. This offset is detectable by TVLA and covers actual leakage
+    # happening during the SHA3 processing. The effect is most easliy visible between loading the
+    # plaintext and appending the padding, i.e., when the target is completely idle and waiting for
+    # the 40 clock cycle timer delay between the RUN and PROCESS command to expire.
+    #
+    # Crafted plaintext value with 4 bits set per byte, and where the Hamming distance between the
+    # first and second 64-bit word is exatly 4 bits per byte.
+    plaintext_fixed = bytearray([0xA5, 0xC3, 0x5A, 0x3C, 0x96, 0x0F, 0x69, 0xF0,
+                                 0xC3, 0xA5, 0x3C, 0x5A, 0x0F, 0x96, 0xF0, 0x69])
```
vrozic, line 616: My preference would be to hide this change behind a parameter loaded from the cfg_file. Something like `capture_cfg["ballanced_fixed_ptext"]`.
This is essentially a debugging feature and we would eventually want to solve the problem in a more robust way.

vrozic (approved):

cw/capture.py
vogelpi: That's a good point @vrozic , I'll try to do something like this. Thanks for your review.

y-srini:
@vogelpi, Thanks for the update. 

Ran TVLA with the balanced plaintexts, on Mask OFF and Mask ON setting. 
Picture below compares the TVLA without and with balanced plaintext (using same bitstreams with which the shift was observed). 

Balanced plaintext definitely seem improve things here, to reduce fake leakages. 

image

abdullahvarici:
I think balancing the plaintext is a good point. Maybe we should also think about this: There is always a possibility of capturing two or more traces for fixed key consecutively since we are determining randomly on runtime whether we are going to send a fixed key input or random. Maybe we should prevent it because some values may not be change if we run the algorithm for the same key causing zero Hamming Weight. But maybe there is already a mechanism like secure wipe. Just an idea.

abdullahvarici (commented):
LGTM!

vogelpi:
FYI: @johannheyszl and I were discussing further about this. After some thinking we thought it would be really interesting that the vertical offset is just introduced by the plaintext loading and not e.g. by previous processing on Ibex or TL-UL transfers (high-fanout nets).

However, it seems that really what is happening. The figure below shows the difference between average traces in the fvsr data test (100k traces) around the plaintext loading. Any offset due to Ibex processing / TL-UL transfers would be to the left of the plaintext loading. As you can see, the difference is nicely centered around 0 there. The offset is only introduced after the peak and then slowly decays back to 0.

Around 1000 samples (50 clock cycles) after the plaintext loading, things are back to normal. However, the time it takes for things to normalize heavily depends on the actual bitstream.

sha3_plaintext_loading

Issue #72, opened 2022-05-24
[tvla] Usability and performance improvements, code cleanup


vogelpi:
With #71 we have moved our TVLA code to the public repo. Besides adding new features, we have so far identified the following things to improve:
- [ ] Use typer (as in `capture.py`) to simplify the handling of input arguments.
- [ ] Allow specifying a command line argument to configure the logger, see https://github.com/lowRISC/ot-sca/pull/71#discussion_r879874972
- [ ] Refactor the main function to simplify it, e.g., by moving chunks of code into helper functions.
- [ ] Parallelize computation using `Ray` (see `ceca.py`).
- [ ] Provide yaml config files to configure specific modes of the TVLA code.
- [ ] Reduce number of command line arguments.
- [ ] Add some form of CI support?

m-temp:
> * [ ]  Allow for analysis of only a part of the trace (e.g. by specifying `start_sample` and `end_sample`)

https://github.com/lowRISC/ot-sca/pull/115/commits/a128197c67ea278bf5c69e21d8440e5fde741cc8 in https://github.com/lowRISC/ot-sca/pull/115 adds this for otbn-vertical. In order to not break backwards compatibility, this feature is deactivated for aes, sha3 and kmac.

vrozic:
#185 Makes trace filtering controllable from the command line. I am therefore marking this item as done.

m-temp:
https://github.com/lowRISC/ot-sca/pull/270 should address the plotting issues with the new database format

wettermo:
https://github.com/lowRISC/ot-sca/pull/270 also adds code in the tvla script to also include metadata in plots of aes specific tests.

vrozic:
Per discussion in #334, I've added an item to correctly calculate `num_traces_used_total` when histograms are added from an input file.

PR #125, opened 2023-02-28
[otbn/tvla] add option to configure the used samples per trace

m-temp:
This option is currently only available for traces with `dtype=np.uint16`
Do we still support traces with `dtype=np.double`?

It is only tested for otbn and rises a runtime error if used for other modes.

See also: https://github.com/lowRISC/ot-sca/issues/72

### DIFF
--- a/cw/tvla.py
+++ b/cw/tvla.py
@@ -488,6 +488,27 @@ def run_tvla(ctx: typer.Context):
         # Amount of tolerable deviation from average during filtering.
         num_sigmas = 3.5
 
+        # Slice of wave file
+        # these options are only tested for otbn
+        if ("sample_start" in cfg and cfg["sample_start"] is not None):
+            sample_start = cfg["sample_start"]
+        else:
+            sample_start = 0
+
+        assert sample_start < len(project.waves[0])
+
+        if ("num_samples" in cfg and cfg["num_samples"] is not None):
+            num_samples = cfg["num_samples"]
+        else:
+            num_samples = len(project.waves[0]) - sample_start
+
+        if (num_samples + sample_start > len(project.waves[0])):
+            log.warning(f"Selected sample window {sample_start} to " +
+                        f"{sample_start+num_samples} is out of range!")
+            num_samples = len(project.waves[0]) - sample_start
+            log.warning(f"Will use samples from {sample_start} " +
+                        f"to {sample_start+num_samples} instead!")
+
         # Overall number of traces, trace start and end indices.
         num_traces_max = len(project.waves)
         if cfg["trace_start"] is None:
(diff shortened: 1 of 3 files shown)


cw/tvla_cfg_otbn.yaml
```diff
@@ -14,3 +14,5 @@ plot_figures: true
 general_test: true
 mode: otbn
 key_len_bytes: 40
+sample_start: null
+num_samples : 4000
```
vrozic, line 18: Could you also add these two lines to `tvla_cfg_template.yaml`, and a comment line `# Only enabled for otbn mode.`?
Eventually, we would like to enable this feature for all modes.

cw/tvla.py
```diff
@@ -488,6 +488,32 @@ def run_tvla(ctx: typer.Context):
...
+        # these options are only tested for otbn
+        if (cfg["mode"] != "otbn" and "sample_start" in cfg and cfg["sample_start"] is not None):
+            raise RuntimeError('Option "sample_start is only supported for otbn!')
+        if (cfg["mode"] != "otbn" and "num_samples" in cfg and cfg["num_samples"] is not None):
+            raise RuntimeError('Option "num_samples is only supported for otbn!')
+
+        if (cfg["mode"] != "otbn" or cfg["sample_start"] is None):
+            sample_start = 0
+        else:
+            sample_start = cfg["sample_start"]
```
vrozic, line 501:
```suggestion
        sample_start = 0
        if ("sample_start" in cfg and cfg["sample_start"] is not None):
            sample_start = cfg["sample_start"]
```

Currently, we will get an error if `sample_start:` is missing in the `.yaml` configuration file. This way it will be automatically set to 0 if not specified.
Similar can be done `num_samples`.
I think that after that we can just remove the `RuntimeError` checks and the feature will work for all modes.

cw/tvla.py
```diff
@@ -488,6 +488,32 @@ def run_tvla(ctx: typer.Context):
...
+
+        assert sample_start < len(project.waves[0])
+
+        if (cfg["mode"] != "otbn" or cfg["num_samples"] is None):
+            num_samples = len(project.waves[0]) - sample_start
+        else:
+            num_samples = cfg["num_samples"]
```
vrozic, line 508:
```suggestion
        num_samples = len(project.waves[0]) - sample_start
        if ("num_samples" in cfg and cfg["num_samples"] is not None):
            num_samples = cfg["num_samples"]
```

cw/tvla_cfg_otbn.yaml
m-temp: yes. good point. Thx for the feedback! I would set both parameters to `null`  as default.

cw/tvla.py
m-temp: Good idea. I simply didn't  have time to check the other modes, but sounds reasonable.

vrozic (approved):

PR #126, opened 2023-02-28
[sca/capture] Add Ctrl-C handler to capture script

wettermo:
In case a huge number of traces is captured, this handler allows the
user, when capturing takes too much time, to abort the capture without
losing previously recorded data.
To use this handler it has to be registered within each of the capture
functions, where a Chipwhisperer project is used to store traces.

### DIFF
--- a/cw/capture.py
+++ b/cw/capture.py
@@ -4,8 +4,11 @@
 # SPDX-License-Identifier: Apache-2.0
 import binascii
 import random
+import signal
+import sys
 import time
 from enum import Enum
+from functools import partial
 from pathlib import Path
 from types import SimpleNamespace
 
@@ -58,6 +61,19 @@ SCOPE_FACTORY = {
     ScopeType.cw: create_cw_segmented,
     ScopeType.waverunner: create_waverunner,
 }
+
+
+def abort_handler(project, sig, frame):
+    """ Handler for ctrl-c keyboard interrupts:
+        Saves capture project before exiting, in case abort is intended.
+        Needs to be registered in every capture function before capture loop.
+        To register handler use:
+        signal.signal(signal.SIGINT, partial(abort_handler, project))
+        where 'project' is the variable for the capture project """
+    if project is not None:
+        print("\nCaught keyboard interrupt -> saving project (traces)...")
+        project.close(save=True)
+    sys.exit(0)
 
 
 # Note: initialize_capture and plot_results are also used by other scripts.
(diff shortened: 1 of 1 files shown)


cw/capture.py
```diff
@@ -17,12 +20,12 @@
...
 from Crypto.Hash import SHA3_256
-from cw_segmented import CwSegmented
-from pyXKCP import pyxkcp
 from tqdm import tqdm
-from waverunner import WaveRunner
 
+from cw_segmented import CwSegmented
+from pyXKCP import pyxkcp
 from util import device, plot
+from waverunner import WaveRunner
```
vogelpi, line 26: Question is this change needed due to lint or is it a remainder from the rebasing maybe?

cw/capture.py
```diff
@@ -258,6 +274,7 @@ def capture_aes_random_batch(ot, ktp, capture_cfg, scope_type):
     num_segments_storage = 1
     # cw and waverunner scopes are supported fot batch capture.
     scope = SCOPE_FACTORY[scope_type](ot, capture_cfg)
+
```
vogelpi, line 277: I think this additioan newline (and the other two below) aren't required. I don't mind them being there to improve readability but then they should be added with the next commit which touches the surroundings of these lines anyway.

vogelpi (commented):
Thanks @wettermo , this looks mostly good.

I have some nits, see comments and two more things:
1. Should we also register the handler inside the `capture_loop` function? This would also enable the feature in non-batch mode for AES, SHA3, KMAC.
2. The commit title of the first commit message is too long, it wraps.

cw/capture.py
wettermo: I had a few problems with isort on my machine, which apparently kind of messed up the imports. I only realized that after I opened the PR, so I had to manually fix it.

cw/capture.py
wettermo: I see, but I don't understand which next commit you mean.

wettermo: Ah sorry, got it!

wettermo:
Thanks @vogelpi for the review!
Good point to also register the handler inside the capture_loop function. I'll do that and I'll also reforge the commits to get rid of the linting/isort issues.

wettermo:
@vogelpi I cleaned up the commits, added the registering of the handler in the capture_loop, and put them together into one commit. Now it should look cleaner.

vogelpi (approved):
Thanks for addressing all comments @wettermo , this looks good to me!

PR #127, opened 2023-03-01
Enhance status print while programming the FPGA

m-temp:
At the moment the "Connecting and loading FPGA..." is only printed after the FPGA is actually programmed.
This is kind of annoying, because programming the FPGA takes some time and nothing is printed during that time.
As this is the very first message to be printed, it is not obvious what's going on and if the python script is running as expected.


vrozic (approved):
LGTM
Thanks @m-temp for changing this

andreaskurth (approved):
Thanks @m-temp!

PR #128, opened 2023-03-03
[otbn/tvla] Add batch mode for OTBN vertical ecc256 keygen capture

wettermo:
This commit adds 

- A batch mode capture function for OTBN vertical ecc256 keygen analysis.
- Additional capture and tvla config files suitable to run with the batch mode (configuration parameters have to be different to the non-batch mode).

The batch mode increases the capture rate, compared to non-batch mode, by approx. 2000% (at least on my FPGA setup). (Before: ca. 20 Traces/s, Now: ca. 400 Traces/s)

First TVLA results look good: (first plot: masks off, 10k traces, second plot: masks on, 40k traces)

otbn_fixed_vs_random_batch_mask_off_10k_traces
otbn_fixed_vs_random_batch_mask_on_40k_traces

Once the corresponding PR for the serial device code (https://github.com/lowRISC/opentitan/pull/17416) is merged I'll open another PR for the built binary.

### DIFF
--- a/cw/capture.py
+++ b/cw/capture.py
@@ -1233,6 +1233,199 @@ def otbn_vertical(ctx: typer.Context,
     capture_end(ctx.obj.cfg)
 
 
+def capture_otbn_vertical_batch(ot, ktp, capture_cfg, scope_type):
+    """A generator for fast capturing otbn vertical (ecc256 keygen) traces.
+    The data collection method is based on the derived test requirements (DTR) for TVLA:
+    https://www.rambus.com/wp-content/uploads/2015/08/TVLA-DTR-with-AES.pdf
+    The measurements are taken by using either fixed or randomly selected seed.
+    In order to simplify the analysis, the first sample has to use fixed seed.
+
+    Args:
+      ot: Initialized OpenTitan target.
+      ktp: Key and plaintext generator.
+      capture_cfg: Capture configuration.
+      scope_type: cw or waverunner as a scope for batch capture.
+    """
+
+    # We need an intuitive, but non default behavor for the kpt.next() interrator.
+    # For backwards compatibility this must be set in the capture config file.
+    # This is a workaroung for https://github.com/lowRISC/ot-sca/issues/116
+    if "use_fixed_key_iter" not in capture_cfg:
+        raise RuntimeError('use_fixed_key_iter not se set!')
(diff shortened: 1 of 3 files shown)


cw/tvla_cfg_otbn_batch..yaml
```diff
@@ -0,0 +1,18 @@
...
+input_file: null
+output_file: null
+number_of_steps: 1
+ttest_step_file: null
+plot_figures: true
+general_test: true
+mode: otbn
+key_len_bytes: 40
+sample_start: null
+num_samples : 3500
```
vrozic, line 18: Do we need a separate configuration file for the batch mode?
It seems that we can continue using `tvla_cfg_otbn.yaml`

cw/capture_otbn_vertical_batch.yaml
```diff
@@ -0,0 +1,37 @@
...
+  output_len_bytes: 40
+  # Samples per trace
+  # Fifo Size is 131070 = max num samples to avoid multiple runs per trace
+  num_samples: 3500
+  # Offset in samples
+  offset: 0
+  # scope gain in db
+  scope_gain: 24
+  num_traces: 40000
+  batch_prng_seed: 6
```
vrozic, line 27: Can we just add `batch_prng_seed:` option to `capture_otbn_vertical.yaml` instead of creating a new configuration file?

wettermo: Sure I can add that. My thought was, because the PLL frequency and number of samples need to be different than in non-batch mode (for non-batch mode 100 MHz is to fast), to add a separate config file for batch mode. Also in batch mode 100 Mhz is hardcoded into cw_segmented.py (see https://github.com/lowRISC/ot-sca/issues/129, @m-temp explained it there). Hence, I thought it might be a good idea to have a separate config file, to be able to run batch mode "out of the box" without having to change capture parameters.

wettermo: I deleted the two additional configuration files, and inserted `batch_prng_seed` into the existing one. Additionally, I added a comment regarding the PLL frequency, that for non-batch mode max. 91 MHz is possible.

vrozic (approved):
To me this looks ready to merge.
Thanks @wettermo for working on this

PR #130, opened 2023-03-07
[otbn/tvla] Add binary for OTBN vertical batch mode

wettermo:
Adds ecc256 keygen serial device binary
Binary compiled from lowrisc OT master after commit: https://github.com/lowRISC/opentitan/commit/894278fb9d7cad111ed715856c0c06a565c366b7


vrozic (approved):
Thanks @wettermo for the update.
I've just tested the new binary and it works well.

wettermo:
Thanks for the review @vrozic ! Glad to hear that it works :)

PR #131, opened 2023-03-07
[tvla] Save capture metadata to CW project file + use it in TVLA plot

wettermo:
This commit adds code, which saves metadata from trace capturing to the CW project file. This data then can be used in the TVLA script without having to hand over the corresponding capture configuration file. Here, this commit adds code to display metadata in the TVLA plot.

In general this makes capture projects as well as TVLA plots more distinguishable, and prevents getting confused with configuration parameters.

For now this is only implemented for OTBN vertical TVLA, but it shouldn't be too much work to add to the other capture modes. A TVLA plot using these modifications looks like this:
otbn_fixed_vs_random

I'll open this as a draft pull request, as this was just some work I did on the side, while working on OTBN vertical capture/TVLA.
We should discuss if we want/need something like this or not, and in case we do, if it still can be improved.

### DIFF
--- a/cw/capture.py
+++ b/cw/capture.py
@@ -7,6 +7,7 @@ import random
 import signal
 import sys
 import time
+from datetime import datetime
 from enum import Enum
 from functools import partial
 from pathlib import Path
@@ -74,6 +75,22 @@ def abort_handler(project, sig, frame):
         print("\nCaught keyboard interrupt -> saving project (traces)...")
         project.close(save=True)
     sys.exit(0)
+
+
+def save_metadata(project, device_cfg, capture_cfg, trigger_cycles, sample_rate):
+    # Save metadata to project file
+    if sample_rate is not None:
+        project.settingsDict['sample_rate'] = sample_rate
+    if device_cfg is not None:
+        for entry in device_cfg:
+            project.settingsDict[entry] = device_cfg[entry]
+    if capture_cfg is not None:
+        for entry in capture_cfg:
+            project.settingsDict[entry] = capture_cfg[entry]
+    # store last number of cycles where the trigger signal was high to metadata
+    if trigger_cycles is not None:
+        project.settingsDict['samples_trigger_high'] = trigger_cycles
+    project.settingsDict['datetime'] = datetime.now().strftime("%m/%d/%Y, %H:%M:%S")
 
 
 # Note: initialize_capture and plot_results are also used by other scripts.
@@ -161,7 +178,7 @@ def capture_init(ctx, num_traces, plot_traces):
     ctx.obj.ot = initialize_capture(cfg["device"], cfg["capture"])
 
 
-def capture_loop(trace_gen, ot, capture_cfg):
+def capture_loop(trace_gen, ot, capture_cfg, device_cfg):
     """Main capture loop.
 
     Args:
@@ -177,6 +194,9 @@ def capture_loop(trace_gen, ot, capture_cfg):
         traces = next(trace_gen)
         check_range(traces.wave, ot.scope.adc.bits_per_sample)
         project.traces.append(traces, dtype=np.uint16)
+
+    sample_rate = int(round(ot.scope.clock.adc_freq, -6))
+    save_metadata(project, device_cfg, capture_cfg, None, sample_rate)
     project.save()
 
 
@@ -214,7 +234,8 @@ def aes_random(ctx: typer.Context,
                plot_traces: int = opt_plot_traces):
     """Capture AES traces from a target that runs the `aes_serial` program."""
     capture_init(ctx, num_traces, plot_traces)
-    capture_loop(capture_aes_random(ctx.obj.ot, ctx.obj.ktp), ctx.obj.ot, ctx.obj.cfg["capture"])
+    capture_loop(capture_aes_random(ctx.obj.ot, ctx.obj.ktp), ctx.obj.ot,
+                 ctx.obj.cfg["capture"], ctx.obj.cfg["device"])
     capture_end(ctx.obj.cfg)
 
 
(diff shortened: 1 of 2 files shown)


cw/capture.py
```diff
@@ -1410,6 +1421,18 @@ def capture_otbn_vertical_batch(ot, ktp, capture_cfg, scope_type):
         project.traces.tm.setTraceSegmentStatus(s, True)
     assert len(project.traces) == capture_cfg["num_traces"]
 
+    # Save metadata to project file
```
abdullahvarici, line 1424: Can we create a function instead of copying the code block?

cw/capture.py
```diff
@@ -1194,6 +1194,17 @@ def capture_otbn_vertical(ot, ktp, fw_bin, pll_frequency, capture_cfg):
...
+    project.settingsDict['pll_frequency'] = pll_frequency
+    project.settingsDict['sample_rate'] = int(round(ot.scope.clock.clkgen_freq, -6))
+    project.settingsDict['masks_off'] = capture_cfg["masks_off"]
+    project.settingsDict['num_samples'] = capture_cfg["num_samples"]
+    project.settingsDict['offset'] = capture_cfg["offset"]
+    project.settingsDict['scope_gain'] = capture_cfg["scope_gain"]
+    project.settingsDict['num_traces'] = capture_cfg["num_traces"]
+    # store last number of cycles where the trigger signal was high to metadata
+    project.settingsDict['samples_trigger_high'] = cycles
+
```
abdullahvarici, line 1207: Maybe we can also add date/time. This may help us to differentiate the results from each other.

cw/tvla.py
```diff
@@ -911,15 +955,57 @@ def run_tvla(ctx: typer.Context):
                     else:
                         plt.close()
         else:
-            #
             c = np.ones(num_samples)
+            xaxs = range(sample_start, sample_start + num_samples)
             fig, axs = plt.subplots(3, sharex=True)
-            axs[0].plot(single_trace, "k")
+            axs[0].set_title("TVLA of " + cfg["project_file"])
```
abdullahvarici, line 961: Maybe we should strip off `projects/` from the header.

cw/tvla.py
```diff
@@ -911,15 +955,57 @@ def run_tvla(ctx: typer.Context):
...
+
+            # Catch case where trigger data isn't saved to project file (e.g. older measurement)
+            try:
+                # Plot trace in different colors, depending on where the trigger is high
```
abdullahvarici, line 965: Should we also do this for t-test graphs?

abdullahvarici (commented):
Thanks for adding this. IMO, this is absolutely useful especially for documentation of our results.

cw/capture.py
wettermo: Good idea!

cw/tvla.py
wettermo: Makes sense, thanks!

m-temp: I think this should be a (hardcoded) parameter. Stripping off is ok as long as we don't have multiple folders.

m-temp (commented):
Thanks @wettermo 
I think this is really useful.  Maybe we could tweak the "trigger low/high" part also for negative offsets, but afaik only AES uses them. So better not overoptimize that. We can always add that if we need it for OTBN.

What we might need later is an option to document the used OTBN-APP.

wettermo:
I added all the comments from @abdullahvarici . Thanks for that! 
Should be ready for review now, and ready to merge afterwards.

cw/capture.py
```diff
@@ -1410,6 +1430,11 @@ def capture_otbn_vertical_batch(ot, ktp, capture_cfg, scope_type):
         project.traces.tm.setTraceSegmentStatus(s, True)
     assert len(project.traces) == capture_cfg["num_traces"]
 
+    # Save metadata to project file
+    pll_frequency = round(scope._scope.clock.clkgen_freq // scope._scope.clock.adc_mul, -6)
+    sample_rate = int(round(scope._scope.clock.clkgen_freq, -6))
+    save_metadata(project, capture_cfg, cycles, pll_frequency, sample_rate)
```
vrozic, line 1436: Could you add `save_metadata` to other capture functions? Currently, if I try to run tvla on e.g. kmac data I get an error because the project doesn't have metadata. 

Note: I think that `cycles` argument will have to be `None` in all function calls other than otbn-vertical

cw/tvla.py
```diff
@@ -911,15 +959,76 @@ def run_tvla(ctx: typer.Context):
...
-            axs[0].plot(single_trace, "k")
-            for i_order in range(num_orders):
-                axs[1 + i_order].plot(ttest_trace[i_order, 0, 0], "k")
-                axs[1 + i_order].plot(c * threshold, "r")
-                axs[1 + i_order].plot(-threshold * c, "r")
-                axs[1 + i_order].set_ylabel('t-test ' + str(i_order + 1))
+            axs[0].set_title("TVLA of " +
+                             (cfg["project_file"]).rsplit('/')[1] + "\n" +
+                             "Captured: " + project.config['ChipWhisperer']
+                             ['General Settings']['datetime'])
```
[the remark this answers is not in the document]
wettermo: You're right this will definitely fail without metadata. I'll add something to handle this.

cw/capture.py
wettermo: Sure! I'll add it to all the other capture functions.
The error you get is probably caused by accessing metadata for the plot title (as you already pointed out below). So I assume that after solving this issue, KMAC TVLA should work again.
But nevertheless, it's a good idea to include this in all the other capture functions.

vrozic (approved):

PR #132, opened 2023-03-22
[otbn/capture] Add fvsr key test for ECC256 keygen

vrozic:
This commit adds the option to capture datasets for fixed vs random key TVLA analysis for ECC256 keygen algorithm on OTBN. This option is hidden behing the test_type parameter. Currently this option is only enabled for normal capture mode in capture_otbn_vertical() function.

### DIFF
--- a/cw/capture.py
+++ b/cw/capture.py
@@ -1167,7 +1167,7 @@ def capture_otbn_vertical(ot, ktp, fw_bin, pll_frequency, capture_cfg, device_cf
     # Initialize some curve-dependent parameters.
     if capture_cfg["curve"] == 'p256':
         curve_order_n = 0xffffffff00000000ffffffffffffffffbce6faada7179e84f3b9cac2fc632551
-        # key_bytes = 256 // 8
+        key_bytes = 256 // 8
         seed_bytes = 320 // 8
     else:
         # TODO: add support for P384
@@ -1182,12 +1182,36 @@ def capture_otbn_vertical(ot, ktp, fw_bin, pll_frequency, capture_cfg, device_cf
         raise ValueError(f'Unexpected mask length: {ktp.textLen()}.\n'
                          f'Hint: set plaintext len={seed_bytes} in the configuration file.')
 
-    # Seed the RNG and generate a random fixed seed for all traces of the
-    # keygen operation.
-    seed_fixed = ktp.next_key()
-    print(f'fixed seed = {seed_fixed.hex()}')
-    if len(seed_fixed) != seed_bytes:
-        raise ValueError(f'Fixed seed length is {len(seed_fixed)}, expected {seed_bytes}')
+    # Generate fixed constants for all traces of the keygen operation.
+
+    if capture_cfg["test_type"] == 'KEY':
+        # In fixed-vs-random KEY mode we use two fixed constants:
+        #    1. C - a 320 bit constant redundancy
+        #    2. fixed_number - a 256 bit number used to derive the fixed key
+        #                      for the fixed set of measurements. Note that in
+        #                      this set, the fixed key is equal to
+        #                      (C + fixed_number) mod curve_order_n
+        C = ktp.next_key()
+        if len(C) != seed_bytes:
+            raise ValueError(f'Fixed seed length is {len(C)}, expected {seed_bytes}')
+        ktp.key_len = key_bytes
+        fixed_number = ktp.next_key()
+        if len(fixed_number) != key_bytes:
+            raise ValueError(f'Fixed key length is {len(fixed_number)}, expected {key_bytes}')
+        ktp.key_len = seed_bytes
+
+        seed_fixed_int = int.from_bytes(C, byteorder='little') + \
+            int.from_bytes(fixed_number, byteorder='little')
+        seed_fixed = seed_fixed_int.to_bytes(seed_bytes, byteorder='little')
+    else:
+        # In fixed-vs-random SEED mode we use only one fixed constant:
+        #    1. seed_fixed - A 320 bit constant used to derive the fixed key
+        #                    for the fixed set of measurements. Note that in
+        #                    this set, the fixed key is equal to:
+        #                    seed_fixed mod curve_order_n
+        seed_fixed = ktp.next_key()
+        if len(seed_fixed) != seed_bytes:
+            raise ValueError(f'Fixed seed length is {len(seed_fixed)}, expected {seed_bytes}')
 
     # Expected key is `seed mod n`, where n is the order of the curve and
     # `seed` is interpreted as little-endian.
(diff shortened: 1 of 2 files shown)


m-temp (changes requested):

cw/capture.py
```diff
@@ -1144,10 +1144,23 @@ def capture_otbn_vertical(ot, ktp, fw_bin, pll_frequency, capture_cfg, device_cf
...
+    if capture_cfg["test_type"] == 'KEY':
+        ktp.key_len = key_bytes
+        key_fixed = ktp.next_key()
+        if len(key_fixed) != key_bytes:
+            raise ValueError(f'Fixed seed length is {len(key_fixed)}, expected {key_bytes}')
+        ktp.key_len = seed_bytes
+
+        intseed = int.from_bytes(C, byteorder='little') + \
+            int.from_bytes(key_fixed, byteorder='little')
+        seed_fixed = intseed.to_bytes(seed_bytes, byteorder='little')
```
wettermo, line 1161: What about the XOR with a random seed1?
And shouldn't this be repeated for every trace with a random seed1?

cw/capture.py
```diff
@@ -1180,6 +1193,15 @@ def capture_otbn_vertical(ot, ktp, fw_bin, pll_frequency, capture_cfg, device_cf
...
             seed_used = seed_fixed
             expected_key = expected_fixed_key
+        elif capture_cfg["test_type"] == 'KEY':
+            # Use a random key with fixed redundancy.
+            ktp.key_len = key_bytes
+            random_key = ktp.next_key()
+            ktp.key_len = seed_bytes
+            intrandom = int.from_bytes(C, byteorder='little') + \
+                int.from_bytes(random_key, byteorder='little')
+            seed_used = intrandom.to_bytes(seed_bytes, byteorder='little')
```
wettermo, line 1203: Similar to above, what about XOR with seed1?
Also, the mask (seed2) is still generated independently from the key. Wasn't the initial idea to calculate seed2 (the mask) in dependency of the fixed/random key and a random seed1, so that we can see leakage of the key and not the seed?

cw/capture.py
vrozic: The xoring is done by ibex in fuction `p256_ecdsa_gen_secret_key`. In the current binary, ibex receives `seed` and `mask`, and then computes `share0 = seed xor mask` and `share1 = mask`.

wettermo: Yes correct, but I thought we want to pre-calculate the mask (seed2) value from the fixed key and a random (for each trace) seed1, and then pass seed1 and seed2 (mask) to the otbn keygen. (similar thought as in the comment below)

cw/capture.py
vrozic: Thanks @wettermo for looking into this.
It is probably confusing that the terminology in the discussion document is not consistent with the variable names. In this code I was trying to be consistent with the variable names that we already use, rather than changing everything to fit the discussion document. Here is how the two terminologies relate (bold font for **discussion document**):
C                   **C**
key_fixed      **fixed_key**
random_key  **random_key**
intseed          **C + fixed_key**
intrandom      **C + random key**
seed_used    **C + random_key** or **C + fixed_key**
mask             **seed1**
computed by ibex **seed2**

I believe that it's functionally correct, but please double check.
I could probably change some variable names to make the code more intuitive.

wettermo: Thanks for pointing this out, probably makes it easier to understand.
But I am still uncertain if this is how we/you intended this. Because seed2 = mask is not computed by Ibex. Ibex just hands over this value to the OTBN as second share. This means that this value is still independent of the fixed/random key.

cw/capture.py
m-temp: hmmm to be honest, I don't have a strong opinion on where to calculate things, but in either case this should be documented in both (python and c) source files.

cw/capture.py
m-temp: We should streamline things here and link the discussion document to the PR and/or even to the commit.
If there is already confusion in the PR, things might get worse later on, when we eventually have a second look on the code.

wettermo: Thanks for pointing this out, might clear up things a bit. 
But still, I thought the mask (and share2) is equal to seed2, while seed1 is equal to the share1 (the seed).

vrozic: Sorry, I was referring to the binary generated from [OT #17502](https://github.com/lowRISC/opentitan/pull/17502) which hasn't yet been merged. Currently the XOR on ibex is happening in function `p256_run_keygen()`

wettermo:
@vrozic and I just had a discussion about my comments. The main reason why there have been ambiguities was that I mixed up seed and mask in the Google-Docs Discussion file, due to inconsistent seed/mask naming.
So this PR is functionally correct!

abdullahvarici (commented):
Thanks @vrozic! I left some comments. Some changes may be done to increase the readability of the code. I also tested it on hardware and had similar results like you had. No leakage with 20k non-batch traces using fvsr-key test.

cw/capture.py
```diff
@@ -1144,10 +1144,23 @@ def capture_otbn_vertical(ot, ktp, fw_bin, pll_frequency, capture_cfg, device_cf
 
     # Seed the RNG and generate a random fixed seed for all traces of the
     # keygen operation.
-    seed_fixed = ktp.next_key()
-    print(f'fixed seed = {seed_fixed.hex()}')
-    if len(seed_fixed) != seed_bytes:
-        raise ValueError(f'Fixed seed length is {len(seed_fixed)}, expected {seed_bytes}')
+
+    C = ktp.next_key()
```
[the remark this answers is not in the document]
vrozic: I've now added a comment to describe this.

cw/capture.py
```diff
@@ -1144,10 +1144,23 @@ def capture_otbn_vertical(ot, ktp, fw_bin, pll_frequency, capture_cfg, device_cf
...
+    if len(C) != seed_bytes:
+        raise ValueError(f'Fixed seed length is {len(C)}, expected {seed_bytes}')
+
+    if capture_cfg["test_type"] == 'KEY':
+        ktp.key_len = key_bytes
+        key_fixed = ktp.next_key()
+        if len(key_fixed) != key_bytes:
+            raise ValueError(f'Fixed seed length is {len(key_fixed)}, expected {key_bytes}')
```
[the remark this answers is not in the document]
vrozic: Fixed

cw/capture.py
vrozic: I've now added a comment to explain what's happening on ibex

cw/capture.py
```diff
@@ -1180,6 +1193,15 @@ def capture_otbn_vertical(ot, ktp, fw_bin, pll_frequency, capture_cfg, device_cf
             # Use the fixed seed.
             seed_used = seed_fixed
             expected_key = expected_fixed_key
+        elif capture_cfg["test_type"] == 'KEY':
```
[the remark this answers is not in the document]
vrozic: Good idea. This is indeed a better way to organise code. I've changed it now.

cw/capture.py
```diff
@@ -1180,6 +1193,15 @@ def capture_otbn_vertical(ot, ktp, fw_bin, pll_frequency, capture_cfg, device_cf
...
+            # Use a random key with fixed redundancy.
+            ktp.key_len = key_bytes
+            random_key = ktp.next_key()
+            ktp.key_len = seed_bytes
+            intrandom = int.from_bytes(C, byteorder='little') + \
```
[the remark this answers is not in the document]
vrozic: Indeed, I've changed it now

cw/capture.py
```diff
@@ -1144,10 +1144,23 @@ def capture_otbn_vertical(ot, ktp, fw_bin, pll_frequency, capture_cfg, device_cf
...
+        raise ValueError(f'Fixed seed length is {len(C)}, expected {seed_bytes}')
+
+    if capture_cfg["test_type"] == 'KEY':
+        ktp.key_len = key_bytes
+        key_fixed = ktp.next_key()
+        if len(key_fixed) != key_bytes:
+            raise ValueError(f'Fixed seed length is {len(key_fixed)}, expected {key_bytes}')
+        ktp.key_len = seed_bytes
+
+        intseed = int.from_bytes(C, byteorder='little') + \
```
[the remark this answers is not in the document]
vrozic: Agreed. It's changed now

cw/capture.py
```diff
@@ -1180,6 +1193,15 @@ def capture_otbn_vertical(ot, ktp, fw_bin, pll_frequency, capture_cfg, device_cf
...
             expected_key = expected_fixed_key
+        elif capture_cfg["test_type"] == 'KEY':
+            # Use a random key with fixed redundancy.
+            ktp.key_len = key_bytes
+            random_key = ktp.next_key()
+            ktp.key_len = seed_bytes
+            intrandom = int.from_bytes(C, byteorder='little') + \
+                int.from_bytes(random_key, byteorder='little')
+            seed_used = intrandom.to_bytes(seed_bytes, byteorder='little')
+            expected_key = int.from_bytes(seed_used, byteorder='little') % curve_order_n
```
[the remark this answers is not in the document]
vrozic: The actual key used is (random_number + C) mod curve_order.
I've changed `random_key` to `random_number` to avoid confusion

abdullahvarici (approved):
Thanks @vrozic! LGTM. I didn't try it on hardware assuming you already tried with both fvsr-key and fvsr-seed tests.

m-temp (approved):

PR #134, opened 2023-03-28
[sca/otbn] Add code to capture and analyze modular inverse OTBN snippet

wettermo:
This PR/commit

- Extends the otbn_vertical capture functions with code to capture traces of modular inversion computation on OTBN.
  -> For this, the OTBN app to be captured has to be configured in the capture config yaml file.
- further, introduces a new TVLA mode 'otbn_modinv', to enable/handle t-test analysis for modular inversion traces.

The corresponding PR in lowrisc/opentitan is https://github.com/lowRISC/opentitan/pull/17734


vogelpi:
@abdullahvarici @vrozic , this and the corresponding OT PR are something you should definitely review.

abdullahvarici:
> @abdullahvarici @vrozic , this and the corresponding OT PR are something you should definitely review.

Actually I was meaning to do that but Jade suggested algorithmic changes on Slack. So I am waiting for @wettermo to implement them at first.

wettermo:
> So I am waiting for @wettermo to implement them at first.

I was ooo the last two days, but I'm currently working on the changes, and I'll let you know when they're ready to review!

wettermo:
I finally finished working on the changes and updated this PR and also the corresponding PR in the OT repository. Should be ready for review now :)

abdullahvarici (commented):
Thanks @wettermo for working on this. I'll also review corresponding PR in Opentitan repository.

cw/capture.py
```diff
@@ -1081,7 +1081,8 @@ def kmac_fvsr_key_batch(ctx: typer.Context,
 
 
 def capture_otbn_vertical(ot, ktp, fw_bin, pll_frequency, capture_cfg, device_cfg):
-    """Capture traces for ECDSA P256/P384 secret key generation.
+    """Capture traces for ECDSA P256/P384 secret key generation
+    and modular inverse computation.
```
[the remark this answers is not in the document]
wettermo: Thanks, I overlooked that. I'll update it.

cw/tvla.py
```diff
@@ -371,17 +371,17 @@ def run_tvla(ctx: typer.Context):
                     force=True,)
 
     if (cfg["mode"] != "kmac" and cfg["mode"] != "aes" and cfg["mode"] != "sha3" and
-            cfg["mode"] != "otbn"):
+            cfg["mode"] != "otbn" and cfg["mode"] != "otbn_modinv"):
```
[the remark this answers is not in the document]
wettermo: Hmm I also had this in mind, but then I thought, depending on how many future OTBN apps will be added, it could be a good idea to only add individual modes if necessary and keep a general `otbn` mode for everything else. Otherwise we might get too many modes, which would make the code even more complex.

cw/capture.py
```diff
@@ -1127,118 +1128,261 @@ def capture_otbn_vertical(ot, ktp, fw_bin, pll_frequency, capture_cfg, device_cf
...
+            f'Curve {capture_cfg["curve"]} is not supported')
 
     # register ctrl-c handler to not lose already recorded traces if measurement is aborted
     signal.signal(signal.SIGINT, partial(abort_handler, project))
 
-    sample_fixed = 1
-    # Loop to collect each power trace
-    for _ in tqdm(range(capture_cfg["num_traces"]), desc='Capturing',
-                  ncols=80):
+    if capture_cfg["app"] == 'keygen':
```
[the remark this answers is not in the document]
wettermo: Thanks @abdullahvarici for pointing this out. This is also something I thought about while writing the code. And yes, there definitely will be merge conflicts with Vladimir's code. But if I modify my PR, and split up the code, merging will get even more complicated. So maybe we should somehow try to solve the merge conflicts between Vladimir and me, and deal with code restructuring afterwards.
Anyway, we should discuss this in the SCA WG meeting, because initially we decided to include everything into one function.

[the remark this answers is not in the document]
wettermo: Just a short follow-up comment to have this documented here as well.
In the SCA WG meeting we decided to deal with the merge conflicts, and then afterwards discuss on how to structure the code.

cw/capture.py
```diff
@@ -1167,118 +1169,269 @@ def capture_otbn_vertical(ot, ktp, fw_bin, pll_frequency, capture_cfg, device_cf
...
+        ot.target.simpleserial_write("a", bytearray([0x01]))
+        time.sleep(0.3)
+
+        # set PRNG seed to ensure same sequence of randoms for split up measurements
+        random.seed(capture_cfg["batch_prng_seed"])
+
+        # set ecc256 fixed key and share inputs
+        # (uncomment the desired fixed shares depending on whether
+        # you want a random fixed key or a hardcoded fixed key)
+        # k_fixed_barray = ktp.next_key()
```
abdullahvarici, line 1321: I think we need to generate 256-bit k but this commented out line is generating a 320-bit k.

cw/capture.py
```diff
@@ -1167,118 +1169,269 @@ def capture_otbn_vertical(ot, ktp, fw_bin, pll_frequency, capture_cfg, device_cf
...
+        k_fixed_barray = bytearray(
```
abdullahvarici, line 1322: Since we already seed the PRNG to generate the same sequence every time we run the capture script, can't we just use above line and delete this?

cw/capture.py
```diff
@@ -1167,118 +1169,269 @@ def capture_otbn_vertical(ot, ktp, fw_bin, pll_frequency, capture_cfg, device_cf
...
+            (0x2648d0d248b70944dfd84c2f85ea5793729112e7cafa50abdf7ef8b7594fa2a1
+             ).to_bytes(key_bytes, 'little'))
+        k_fixed = int.from_bytes(k_fixed_barray, byteorder='little')
```
abdullahvarici, line 1325: k_fixed should be set here, outside of the loop. But k0 and k1 should be randomly generated and set in the for loop. I think, we need to move below lines under `if sample_fixed:` in the for loop.

abdullahvarici (commented):
Thanks @wettermo working on this. I left some feedbacks. I am looking forward to eliminate all the leakages :)

cw/capture.py
wettermo: Ups, you're right! I'll fix it.

cw/capture.py
wettermo: Correct, thanks Abdullah! Missed this.

cw/capture.py
```diff
@@ -1167,118 +1169,269 @@ def capture_otbn_vertical(ot, ktp, fw_bin, pll_frequency, capture_cfg, device_cf
...
+        # generate two random shares
+        input_k0_fixed = ktp.next_key()
+        input_k1_fixed = ktp.next_key()
+        k0_fixed = int.from_bytes(input_k0_fixed, byteorder='little')
+        k1_fixed = int.from_bytes(input_k1_fixed, byteorder='little')
+        # adapt share k1 so that k = (k0 + k1) mod n
+        k_tmp = (k0_fixed + k1_fixed) % curve_order_n
+        k_tmp_diff = k_fixed - k_tmp
```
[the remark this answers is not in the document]
wettermo: Hmm I don't think this is necessary, because a negative difference, in my opinion, is not problematic. Actually, when k_tmp is bigger than k_fixed we need to subtract the difference from one of the shares to get k_fixed as a result from the modulo operation of the share sum. If the difference would be positive, even if k_tmp is bigger than k_fixed, in my opinion this would not lead to k_fixed as result from the modulo operation of the share sum.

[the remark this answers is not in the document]
abdullahvarici: But if the k_tmp_diff is negative then there is a slight chance k1_fixed may become negative, right?

cw/capture.py
```diff
@@ -1167,118 +1169,269 @@ def capture_otbn_vertical(ot, ktp, fw_bin, pll_frequency, capture_cfg, device_cf
...
+        k1_fixed = k1_fixed + k_tmp_diff
```
abdullahvarici, line 1334: Also we need to check if k1_fixed is still less than 2**320 after this addition. If it is 321 bits then we should substract mod value from it, I think.

cw/capture.py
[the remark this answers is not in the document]
wettermo: Ahh yes true, didn't have in mind, that theoretically k1_fixed can be for example 320-bit zero.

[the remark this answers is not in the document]
abdullahvarici: Exactly!

cw/capture.py
wettermo: You're right, thanks!

cw/capture.py
```diff
@@ -1167,118 +1169,266 @@ def capture_otbn_vertical(ot, ktp, fw_bin, pll_frequency, capture_cfg, device_cf
...
+                f'Hint: set plaintext len={seed_bytes} in the configuration file.'
+            )
 
-        if reset_firmware:
-            ot.program_target(fw_bin, pll_frequency)
+        # select the otbn app on the device (0 -> keygen, 1 -> modinv)
+        ot.target.simpleserial_write("a", bytearray([0x00]))
+        time.sleep(0.3)
+
+        # Seed the RNG and generate a random fixed seed for all traces of the
```
abdullahvarici, line 1201: Aren't we need to seed RNG before generating a fixed seed value?

wettermo: Yes of course!
Don't know why this is missing here. I'll fix it!!
Thanks for pointing this out!

wettermo:
I updated the PR with my changes to solve the merge conflicts that appeared after merging Vladimir's PR (https://github.com/lowRISC/ot-sca/pull/132).
@abdullahvarici Would you mind giving it another review, so that finally it can be merged? :)

m-temp:
imho `cw/tvla_cfg_otbn_modinv.yaml` and `cw/tvla_cfg_otbn_modinv.yaml` are missing a newlines at the eof.

cw/capture.py
```diff
@@ -1169,168 +1171,315 @@ def capture_otbn_vertical(ot, ktp, fw_bin, pll_frequency, capture_cfg, device_cf
...
+            C = ktp.next_key()
+            if len(C) != seed_bytes:
+                raise ValueError(f'Fixed seed length is {len(C)}, expected {seed_bytes}')
+            ktp.key_len = key_bytes
+            fixed_number = ktp.next_key()
+            if len(fixed_number) != key_bytes:
+                raise ValueError(f'Fixed key length is {len(fixed_number)}, expected {key_bytes}')
+            ktp.key_len = seed_bytes
+
+            seed_fixed_int = int.from_bytes(C, byteorder='little') + \
```
abdullahvarici, line 1220: This might become a 321-bit number, right? Should we substract n from it if it becomes a 321-bit number?

abdullahvarici (approved):
Thanks @wettermo. Looks good to me. But I think @vrozic should also review this before merging.

abdullahvarici:
I also tried this PR locally and saw that keygen-fvsr-key-batch/nonbatch, keygen-fvsr-seed-batch/nonbatch and modinv-fvsr-key-nonbatch are all working.

cw/capture.py
```diff
@@ -1169,168 +1171,315 @@ def capture_otbn_vertical(ot, ktp, fw_bin, pll_frequency, capture_cfg, device_cf
...
-        if capture_cfg["masks_off"] is True:
-            # Use a constant mask for each trace
-            mask = bytearray(capture_cfg["plain_text_len_bytes"])  # all zeros
-        else:
-            # Generate a new random mask for each trace.
-            mask = ktp.next_text()
-
-        tqdm.write("Starting new trace....")
-        tqdm.write(f'mask   = {mask.hex()}')
+        # Seed the RNG and generate a random fixed seed for all traces of the keygen operation
```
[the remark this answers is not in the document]
wettermo: Thanks, fixed!

cw/capture.py
wettermo: Theoretically yes I would say. Might be a good idea to subtract n.
But maybe it would be a good idea to do this in a separate PR?

vogelpi:
Hi @wettermo ,
I see that this PR has been approved and the corresponding OT PRs have been merged already. Do you want to add the updated binary into this PR?

wettermo:
Of course I'll add them. Thanks @vogelpi !
But I would like to wait with merging this PR until @vrozic gives his final go.

cw/capture.py
vrozic: Yes, this seems to be a pre-existing issue.
However, the only way an overflow could happen is if C > 2^320 - 2^64, which is probably not the constant we want to use.
If we decide to add a check for this I would prefer to raise an error rather than subtract `n`. Either way, this PR doesn't need to change it.

PR #136, opened 2023-03-31
[capture] Add force programming of FPGA over CLI

abdullahvarici:
This feature is especially necessary when accessing FPGAs connected to remote server and one cannot power-off and on the FPGA board to trigger bitstream programming.

### DIFF
--- a/cw/capture.py
+++ b/cw/capture.py
@@ -38,7 +38,8 @@ app = typer.Typer(add_completion=False)
 # To be able to define subcommands for the "capture" command.
 app_capture = typer.Typer()
 app.add_typer(app_capture, name="capture", help="Capture traces for SCA")
-# Shared options for "capture aes" and "capture kmac".
+# Shared options for capture commands
+opt_force_program_bitstream = typer.Option(None, help=("Force program FPGA with the bitstream."))
 opt_num_traces = typer.Option(None, help="Number of traces to capture.")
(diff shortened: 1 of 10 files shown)


cw/capture.py
```diff
@@ -156,9 +158,12 @@ def init(ctx: typer.Context):
     initialize_capture(ctx.obj.cfg["device"], ctx.obj.cfg["capture"])
 
 
-def capture_init(ctx, num_traces, plot_traces):
+def capture_init(ctx, force_program_bitstream, num_traces, plot_traces):
     """Initializes the user data stored in the context and programs the target."""
     cfg = ctx.obj.cfg
+    if force_program_bitstream is not None:
```
vrozic, line 164: It is perhaps better to change this to `if force_program_bitstream:`

vrozic (approved):
Thanks @abdullahvarici for implementing this feature.
I've just tested it and it works OK.
The PR looks good to me.

cw/capture.py
abdullahvarici: But if I do that, it won't be possible to overwrite this option in CLI with `--no-force-program-bitstream` when it is already defined as True in yaml file.

vrozic: You're right. it wouldn't be possible to override this setting in command line.
OK, then i think this is ready to merge

vogelpi (approved):
Nice works, thanks for adding this @abdullahvarici !

Issue #83, opened 2022-06-20
Debug aes-fvsr-key-batch timing issues


vogelpi:
PR #82 added the corresponding target binaries to support the aes-fvsr-key-batch command. For most of us this seems to work just fine but I saw lots of errors like:
```
Fixed key: b'811e3731b0120a7842781e22b25cddf9'
Connected to ChipWhisperer (num_samples: 1200, num_samples_actual: 1200, num_segments_actual: 1)
Capturing:  16%|███▋                   | 162/1000 [00:00<00:01, 492.97 traces/s]
Traceback (most recent call last):
  File "./capture.py", line 771, in <module>
    app()
 ...
  File "./capture.py", line 502, in aes_fvsr_key_batch
    capture_aes_fvsr_key_batch(
  File "./capture.py", line 460, in capture_aes_fvsr_key_batch
    check_ciphertext(ot, expected_last_ciphertext, 4)
  File "./capture.py", line 221, in check_ciphertext
    assert actual_last_ciphertext == expected_last_ciphertext[0:ciphertext_len], (
AssertionError: Incorrect encryption result!
actual: bytearray(b'^\xdc\xe8{')
expected: [ 85 154 226   3  30  12  31 147  54 197 151  35   4 134  80  98]
```

In words, there are sometimes mismatches between the expected and received ciphertexts. Sometimes, the failure occurs for the first batch, sometimes for a later batch, sometimes it doesn't occur at all. The failures seem to depend on timing (adding some sleep command on the Python side seem to help), Husky firmware (the latest firmware seems to be more affected), and maybe also USB connection setup (docking station, hub, laptop directly).

We should root cause and fix the problem. Otherwise we can't reliably do long running captures. Imagine we collect 10 Mio traces  and get such a failure after 2 hours - all traces will be lost.

abdullahvarici:
I have investigated the issue and I guess I found the root cause. 

We have two consecutive serial write command in 'capture_aes_fvsr_key_batch()' function in 'capture.py' script:
`ot.target.simpleserial_write("s", capture_cfg["batch_prng_seed"].to_bytes(4, "little"))
 ot.target.simpleserial_write("t", key_fixed)`
In the device firmware 'aes_serial.c', we are polling incoming packets and call the relevant function: `while (true) {
    simple_serial_process_packet();
  }`.

[...]

PR #137, opened 2023-04-14
[aes] Fix aes_fvsr_key_batch on CW305

vogelpi:
Previously, the commands for seeding the batch PRNG and seeting the fixed key seemed to interfere resulting in the target implemented on the CW305 getting completely out of sync.

### DIFF
--- a/cw/capture.py
+++ b/cw/capture.py
@@ -449,6 +449,10 @@ def capture_aes_fvsr_key_batch(ot, ktp, capture_cfg, scope_type, gen_ciphertexts
     # Seed the target's PRNG
     ot.target.simpleserial_write("s", capture_cfg["batch_prng_seed"].to_bytes(4, "little"))
 
+    # Set and transfer the fixed key.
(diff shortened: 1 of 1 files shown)


abdullahvarici (approved):
LGTM. Thanks Pirmin but we can think this as a temporary and quick fix. But at the end we should implement [Colin's suggestion](https://github.com/lowRISC/ot-sca/issues/83#issuecomment-1221042467) but it also needs some modification on OTBN binary side.

vogelpi:
> LGTM. Thanks Pirmin but we can think this as a temporary and quick fix. But at the end we should implement [Colin's suggestion](https://github.com/lowRISC/ot-sca/issues/83#issuecomment-1221042467) but it also needs some modification on OTBN binary side.

Thanks for your review @abdullahvarici ! Ah, that's were this issue is :-) I vaguely remembered something but forgot to check. I agree that we should investigate this and come up with a real/more reliable fix.

PR #138, opened 2023-04-14
[aes] Add separate SW LFSR for key masking

vogelpi:
This requires the AES target binaries to be updated as well. I'll publish a PR for the changes to `aes_serial.c` upstream and then add the updated binaries to this PR as well.


vogelpi:
Here is the corresponding OT PR: https://github.com/lowRISC/opentitan/pull/17987

abdullahvarici (approved):
Thanks @vogelpi!

vogelpi:
Update: the upstream PR has been merged a long time ago and I've finally managed to re-generate and test all AES bitstreams and binaries. I've now added this to this PR and will merge it once CI passes.

PR #139, opened 2023-04-27
[capture] Set PRNG seed prior to setting fixed seed

abdullahvarici:
This is to have the same input sequence every time running this function with the same batch_prng_seed.

Corresponding PR in Opentitan repo: https://github.com/lowRISC/opentitan/pull/18210


wettermo (approved):
Thanks @abdullahvarici, this change makes sense.

PR #142, opened 2023-06-14
[capture] Replacing hardcoded value with pll_frequency

### LINKED ISSUE #129: scope.clock.clkgen_freq vs pll_frequency in batch modes
While writing the code for the otbn capture and helping debugging the otbn-batch mode, @wettermo and I stumbled across the following lines
https://github.com/lowRISC/ot-sca/blob/111c14089e5df384ea77da8c8c6f585e29b712e6/cw/cw_segmented.py#L175-L187

Can someone explain to me, why `self._scope.clock.clkgen_freq = 100000000` is hardcoded there? Is this just a leftover or is there a particular reason? I'd expected thatt, `self._scope.clock.clkgen_freq = pll_frequency` like in https://github.com/lowRISC/ot-sca/blob/111c14089e5df384ea77da8c8c6f585e29b712e6/cw/util/device.py#L130

While writing, I just discovered that @bilgiday had changed this behavior in `cw/util/device.py` in https://github.com/lowRISC/ot-sca/commit/f51c6ec662334e3940169e449946ecfa6e958770

So, may I assume, we could change this also in the `cw_segmented.py`, or is this a bad idea for some reason?

vrozic:
This commit replaces the hardcoded value of 100000000 for pll_frequency in cw_segmented with a value loaded from the config file.

This change closes #129

### DIFF
--- a/cw/capture.py
+++ b/cw/capture.py
@@ -51,12 +51,13 @@ def create_waverunner(ot, capture_cfg):
     return WaveRunner(capture_cfg["waverunner_ip"])
 
 
-def create_cw_segmented(ot, capture_cfg):
+def create_cw_segmented(ot, capture_cfg, device_cfg):
     """Create CwSegmented object to be used for batch capture."""
     return CwSegmented(num_samples=capture_cfg["num_samples"],
                        offset=capture_cfg["offset"],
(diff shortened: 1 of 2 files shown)


vogelpi (approved):
LGTM, thanks @vrozic !

PR #143, opened 2023-06-14
[capture] Removes variable reset_firmware

### LINKED ISSUE #105: [ecdsa] Don't reset target between traces
It seems the target is currently reset between capturing different traces or trace segments which leads to very low performance as the binary needs to be transferred over SPI for every segment. We should look into changing the capture setup to not require a reset of the target between segments.

vrozic:
Code-cleanup for capture.py.
Removes variable reset_firmware as discussed in issue #105 

Closes #105

### DIFF
--- a/cw/capture.py
+++ b/cw/capture.py
@@ -1145,10 +1145,6 @@ def capture_otbn_vertical(ot, ktp, fw_bin, pll_frequency, capture_cfg, device_cf
     if capture_cfg["use_fixed_key_iter"] is not False:
         raise RuntimeError('use_fixed_key_iter must be set to false!')
 
-    # This determines if we want to reprogram the firmware
-    # before every sign operation.The default value is false.
-    reset_firmware = False
-
     # OTBN operations are long. CW-Husky can store only 131070 samples
     # in the non-stream mode.
     fifo_size = 131070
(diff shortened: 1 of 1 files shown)


vogelpi (approved):
LGTM, thanks @vrozic !

cw/capture.py
```diff
@@ -1772,17 +1762,10 @@ def capture_ecdsa_sections(ot, fw_bin, pll_frequency, num_sections, secret_k, pr
...
     """
 
-    # This determines if we want to reprogram the firmware
-    # before every sign operation.The default value is false.
-    reset_firmware = False
-
     # Create a temporary buffer to keep the collected sections
     tmp_buffer = np.array([])
     for ii in range(num_sections):
         # Reflash the firmware and set the CW310 pll frequency
```
vogelpi, line 1768: I think also this comment needs to be removed. Similar occurrence below.

cw/capture.py
```diff
@@ -2133,9 +2112,6 @@ def capture_ecdsa_stream(ot, fw_bin, pll_frequency, capture_cfg):
         # Create an arrey to keep the traces
         waves = np.array([])
         # Reflash the firmware and set the CW310 pll frequency
```
vogelpi, line 2114: See here

cw/capture.py
vrozic: Agreed, I've removed this comment in both places with the latest change

PR #144, opened 2023-06-23
[tvla] Bug Fix

vrozic:
A bug was introduced to tvla in PR #134 when removing references to mode["otbn_modinv"].
It seems that a part of the condition was not deleted in one of the if branches, which results in an error when running tvla for KMAC.
This commit fixes the condition.


vrozic:
> Oh no... that was my fault, sorry for that. I just missed this one. Thanks for fixing it @vrozic !!

No worries. This bug actually helped me identify a problem with our CI, so it's good that it happened.
Thanks for the review.

PR #145, opened 2023-06-26
[ci] Adds a CI test - opening KMAC project

vrozic:
This test checks if the KMAC project can be succesfully opened by the tvla analysis script.


vrozic:
Currently, CI tests for `tvla.py` are only covering the mode when input data is loaded from the histogram file. About 25 % of the tvla code is in the `if` branch which is completely skipped in this case. This has lead to at least one bug being introduced and unnoticed for a long time.
To improve the code coverage, I've added a test that loads traces from a project rather than histograms. This should hopefully help us avoid new bugs in the future.

I plan to do some `tvla.py` code clean-up and restructuring in the near future; this is to ensure that things still work afterwards.

cw/tvla.py
```diff
@@ -776,6 +776,8 @@ def run_tvla(ctx: typer.Context):
                                             x_axis)
                 for i in range(0, num_samples, sample_step_ttest))
             ttest_trace = np.concatenate((ttest_trace[:]), axis=3)
+            log.info("Saving T-test")
+            np.save('tmp/ttest.npy', ttest_trace)
```
vogelpi, line 780: Oh, now I understand the problem. In CI we are doing a first call to the tvla script (which previously wouldn't update the `tmp/ttest.npy`)  including leakage detection. Since the tvla script always succeeds and we didn't want to parse  the log file to extract the pass/fail result, we added a separate leakage detection function that loads `tmp/ttest.npy`. And this was now using the old result from the previous CI test which is why you had to add this `np.save()` command here.

vogelpi: An alternative solution to this would have been to call the tvla script for this particular new test with `--save_to_disk` and then adapt the CI custom leakage detection function to also operate on the `ttest-step.npy` file).

vogelpi (approved):
Thanks @vrozic , this required some serious debugging. Glad to see that you debugged the issue and found a solution. LGTM!

cw/tvla.py
vrozic: > An alternative solution to this would have been to call the tvla script for this particular new test with `--save_to_disk` and then adapt the CI custom leakage detection function to also operate on the `ttest-step.npy` file).

Good idea. I will look into this when I implement tests for the step feature. Unfortunately, it seems that, at this moment, we can generate `ttest-step.npy` only if we are also saving leakage histograms in the output file (another strange feature of `tvla.py`). I'll add this to the tracking issue #72 to clean-up later.

PR #147, opened 2023-07-27
[capture] Replace pyxkcp.kmac128 with pycryptodome

vrozic:
Capture setup for kmac was developed a long time ago, before pycryprodome supported KMAC128. As a temporary solution we used pyXKCP to verify the computation result.
In the meantime KMAC128 was added to the latest version of pycryptodome.

This PR replaces calls to pyxkcp.kmac128 with the pycryptodome calls.

I've tested the modified capture functions on FPGA and it all works properly.


vrozic:
> . Should we also add a minimum version to pycryptodome,

Yes of course. I've now added minimal version 3.12.0
Thanks @m-temp  for catching this.


> we should deal with those leftovers in a second commit/PR/issue.
Indeed, I've added a second commit that removes the files.

m-temp (approved):
LGTM. I've tested the changes locally and they are still running :)
Thx @vrozic  for taking care of this!

PR #148, opened 2023-08-09
[tvla] clean up

vrozic:
This PR cleans up tvla code:
1. Renames histogram related arguments
2. Fixes typos and code duplication
3. Refactors code related to saving t-test results

### DIFF
--- a/cw/tvla.py
+++ b/cw/tvla.py
@@ -424,9 +424,9 @@ def run_tvla(ctx: typer.Context):
         save_to_disk_trace = False
         save_to_disk_leakage = False
 
-    if cfg["input_file"] is not None:
+    if cfg["input_histogram_file"] is not None:
         # Load previously generated histograms.
-        histograms_file = np.load(cfg["input_file"])
+        histograms_file = np.load(cfg["input_histogram_file"])
         histograms_in = histograms_file['histograms']
         single_trace = histograms_file['single_trace']
         num_samples = histograms_in.shape[3]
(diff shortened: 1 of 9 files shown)


test/tvla_test.py
```diff
@@ -31,22 +31,22 @@ def ttest_significant(ttest_trace) -> bool:
 def test_general_nonleaking_project():
     project_path = TestDataPath('tvla_general/ci_opentitan_simple_kmac.cwp')
     tvla = TvlaCmd(Args(['--project-file', str(project_path),
-                        '--mode', 'kmac', 'run-tvla'])).run()
+                        '--mode', 'kmac', '--save-to-disk', 'run-tvla'])).run()
```
vogelpi, line 34: I am not sure if we should add a specific argument to only save the ttest[-step] file. Because this argument here will also save the trace file to disk (a dense copy of the chipwhisperer traces). It might not be ideal for CI due to I/O limitations even though the trace sets here are small.

For non-CI applications, i.e., when large trace sets are analyzed, this will store several GBytes of data to disk which is slow. For this reason I originally only enabled `save_to_disk_trace` and `save_to_disk_trace` when number of steps is 1. That should maybe be reworked, can be done in another PR of course!

vogelpi (approved):
LGTM, thanks for working on this @vrozic .

Also, thanks for splitting your commits into well described atomic commits. This is very nice for reviewing :+1:

cw/tvla.py
```diff
@@ -827,6 +814,19 @@ def run_tvla(ctx: typer.Context):
             single_trace = np.load(single_trace_file)
             assert num_samples == single_trace.shape[0]
 
+    if save_to_disk_ttest:
+        if num_steps > 1:
```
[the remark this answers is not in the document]
vrozic: Good point. I've changed it now so that `save_to_disk_ttest` is generated as:
`save_to_disk_ttest = cfg["save_to_disk_ttest"] and (cfg["ttest_step_file"] is None)`

test/tvla_test.py
vrozic: I agree.
For now I've added `--save-to-disk-ttest` option and updated issue #72 to do the same for `--save-to-disk-leakage` and `--save-to-disk-trace`.

vrozic:
Thanks @vogelpi for the review.
I have addressed the comments in the latest push.

vogelpi:
Perfect, thanks @vrozic . Feel free to go ahead and merge this.

Issue #149, opened 2023-08-10
[objs] update ECC384 binaries


vrozic:
OT [PR 19380](https://github.com/lowRISC/opentitan/pull/19380) updated `ecc384_serial.c` and `p384_ecdsa_sca.s` to enable trace capturing with the latest setup.

Binaries should be updated.

vogelpi:
Hey @vrozic , @bilgiday, @wettermo ,
I've updated lowRISC/OpenTitan#19380 and intend to merge it once CI passes again. Do you maybe want to try it out again on the board?

vrozic:
#204 Updates this binary and should close this issue.

PR #150, opened 2023-08-10
Add simple test script to test all modes supported by capture.py

vogelpi:
This PR adds a very simple script to test all capture modes currently supported by `capture.py`. What it does is running the capture commands and copy the trace sample figures under `tmp` for visual inspection. This is useful for a couple of things we need to do soon:
1. Updating ChipWhisperer API
2. Adding CI support
3. Refactoring `capture.py`

The script is not beautiful nor smart but it does what it needs. We can improve / rewrite as we do the things above.

### DIFF
--- a/cw/test_capture.sh
+++ b/cw/test_capture.sh
@@ -1,0 +1,126 @@
+#!/bin/bash
+# Copyright lowRISC contributors.
+# Licensed under the Apache License, Version 2.0, see LICENSE for details.
+# SPDX-License-Identifier: Apache-2.0
+
+# Simple script to test all relevant capture modes supported by capture.py.
+# To test capture modes on the CW305 board, type
+#
+# ./test_capture.sh cw305
+#
+# For the CW310, simply type
+#
+# ./test_capture.sh
+#
+
+set -e
+set -o pipefail
+
+error () {
+    echo >&2 "$@"
+    exit 1
+}
+
+# Input arguments.
+# Only one argument is supported to specify the FPGA board.
+if [ $# -lt 1 ]; then
+  BOARD=cw310
+else
+  BOARD=$1
+fi
+if `[ "${BOARD}" != "cw305" ] && [ "${BOARD}" != "cw310" ]`; then
+  error "Board ${BOARD} not supported"
+fi
+
+# Create results folder.
+mkdir -p tmp
+
+# AES
+MODE="aes"
+declare -A aes_test_list
+aes_test_list["aes-random"]=100
+aes_test_list["aes-random-batch"]=1000
+aes_test_list["aes-fvsr-key"]=100
+aes_test_list["aes-fvsr-key-batch"]=1000
+
+ARGS="--force-program-bitstream"
+for test in ${!aes_test_list[@]}; do
+  echo Testing ${test} on ${BOARD}
+  echo Testing ${test} on ${BOARD} - `date` >> "tmp/${BOARD}_test_capture.log"
+  NUM_TRACES=${aes_test_list[${test}]}
+  ./capture.py --cfg-file capture_${MODE}_${BOARD}.yaml capture ${test} \
+      --num-traces ${NUM_TRACES} ${ARGS} &>> "tmp/${BOARD}_test_capture.log"
+  mv projects/sample_traces_${MODE}.html tmp/${BOARD}_${test}_traces.html
+  ARGS=""
+done
+
+if [ ${BOARD} == "cw310" ]; then
+  # KMAC
+  MODE="kmac"
+  declare -A kmac_test_list
+  kmac_test_list["kmac-random"]=100
+  kmac_test_list["kmac-fvsr-key"]=100
+  kmac_test_list["kmac-fvsr-key-batch"]=1000
+
+  ARGS="--force-program-bitstream"
+  for test in ${!kmac_test_list[@]}; do
+    echo Testing ${test} on ${BOARD}
+    echo Testing ${test} on ${BOARD} - `date` >> "tmp/${BOARD}_test_capture.log"
+    NUM_TRACES=${kmac_test_list[${test}]}
+    ./capture.py --cfg-file capture_${MODE}_${BOARD}.yaml capture ${test} \
+        --num-traces ${NUM_TRACES} ${ARGS} &>> "tmp/${BOARD}_test_capture.log"
+    mv projects/sample_traces_${MODE}.html tmp/${BOARD}_${test}_traces.html
+    ARGS=""
+  done
+
+  # SHA3
+  MODE="sha3"
+  declare -A sha3_test_list
+  sha3_test_list["sha3-fvsr-data"]=100
+  sha3_test_list["sha3-fvsr-data-batch"]=1000
+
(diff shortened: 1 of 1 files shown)


cw/test_capture.sh
```diff
@@ -0,0 +1,106 @@
...
+      ARGS=""
+    fi
+    NUM_TRACES=${sha3_test_list[${test}]}
+    ./capture.py --cfg-file capture_${MODE}_${BOARD}.yaml capture ${test} \
+        --num-traces ${NUM_TRACES} ${ARGS} &>> "tmp/${BOARD}_test_capture.log"
+    mv projects/sample_traces_${MODE}.html tmp/${BOARD}_${test}_traces.html
+  done
+
+  # OTBN
+  # TODO!
```
vogelpi, line 103: meet.google.com/aqf-pkqy-tjm

vogelpi, line 103: @wettermo and @vrozic, would you mind providing me the OTBN commands we're currently supporting? I don't have a good overview of what's working and known broken atm. Thanks!

cw/test_capture.sh
```diff
@@ -0,0 +1,106 @@
...
+aes_test_list["aes-fvsr-key-batch"]=1000
+
+for test in ${!aes_test_list[@]}; do
+  echo Testing ${test} on ${BOARD}
+  echo Testing ${test} on ${BOARD} - `date` >> "tmp/${BOARD}_test_capture.log"
+  if [ "${test}" = "aes-random" ]; then
+    ARGS="--force-program-bitstream"
+  else
+    ARGS=""
+  fi
```
vrozic, line 53: Tests are not always executed in the order they is listed here. Consequently, argument `--force-program-bitstream` is not always set for the first test that is executed. For this reason, some tests end up running using a wrong bitstream.
I would fix this by moving  `ARGS="--force-program-bitstream"` before the for loop and `ARGS=""` at the end of the loop body.

cw/test_capture.sh
```diff
@@ -0,0 +1,106 @@
...
+    if [ "${test}" = "kmac-random" ]; then
+      ARGS="--force-program-bitstream"
+    else
+      ARGS=""
+    fi
+    NUM_TRACES=${kmac_test_list[${test}]}
+    ./capture.py --cfg-file capture_${MODE}_${BOARD}.yaml capture ${test} \
+        --num-traces ${NUM_TRACES} ${ARGS} &>> "tmp/${BOARD}_test_capture.log"
+    mv projects/sample_traces_${MODE}.html tmp/${BOARD}_${test}_traces.html
+  done
```
vrozic, line 80: Same as above

cw/test_capture.sh
```diff
@@ -0,0 +1,106 @@
...
+    if [ "${test}" = "sha3-fvsr-data" ]; then
+      ARGS="--force-program-bitstream"
+    else
+      ARGS=""
+    fi
+    NUM_TRACES=${sha3_test_list[${test}]}
+    ./capture.py --cfg-file capture_${MODE}_${BOARD}.yaml capture ${test} \
+        --num-traces ${NUM_TRACES} ${ARGS} &>> "tmp/${BOARD}_test_capture.log"
+    mv projects/sample_traces_${MODE}.html tmp/${BOARD}_${test}_traces.html
+  done
```
vrozic, line 100: same as above

cw/test_capture.sh
wettermo: @vogelpi basically there are just two otbn commands:
otbn_test_list["otbn-vertical"]=100
otbn_test_list["otbn-vertical-batch"]=1000
The differentiation between keygen and modinv is done through an "app" parameter in the relevant capture config files

vrozic, line 103:
```suggestion
  # OTBN keygen
  MODE="otbn_vertical_keygen"
  declare -A otbn_keygen_test_list
  otbn_keygen_test_list["otbn-vertical"]=100
  otbn_keygen_test_list["otbn-vertical-batch"]=1000

  ARGS="--force-program-bitstream"
  for test in ${!otbn_keygen_test_list[@]}; do
    echo Testing ${test} on ${BOARD}
    echo Testing ${test} on ${BOARD} - `date` >> "tmp/${BOARD}_test_capture.log"
    NUM_TRACES=${otbn_keygen_test_list[${test}]}
    ./capture.py --cfg-file capture_${MODE}.yaml capture ${test} \
        --num-traces ${NUM_TRACES} ${ARGS} &>> "tmp/${BOARD}_test_capture.log"
    mv projects/sample_traces_ecdsa_keygen.html tmp/${BOARD}_${test}_traces.html
    ARGS=""
  done

  # OTBN modinv
  MODE="otbn_vertical_modinv"
  declare -A otbn_modinv_test_list
  otbn_modinv_test_list["otbn-vertical"]=100

  for test in ${!otbn_modinv_test_list[@]}; do
    echo Testing ${test} on ${BOARD}
    echo Testing ${test} on ${BOARD} - `date` >> "tmp/${BOARD}_test_capture.log"
    NUM_TRACES=${otbn_modinv_test_list[${test}]}
    ./capture.py --cfg-file capture_${MODE}.yaml capture ${test} \
        --num-traces ${NUM_TRACES} ${ARGS} &>> "tmp/${BOARD}_test_capture.log"
    mv projects/sample_traces_ecdsa_modinv.html tmp/${BOARD}_${test}_traces.html
  done
```

Something like this works well for testing otbn vertical keygen and modinv.
A couple of caveats:

1. `pll_frequency` in `capture_otbn_vertical_modinv.yaml` needs to be reduced to 91 MHz
2. parameter `test_type:` (KEY or SEED) cannot be overridden through command line. This is currently only testing for `test_type: KEY`.
3. OTBN commands don't follow the same naming strategy used in the rest of the script. As a result we end up with two tests named `otbn_vertical`. We should come up with a convention for naming capture commands and modes. I'll open an issue for this.
4. This doesn't include capture of ECDSA256 and ECDSA384. I suggest to add those later after we close #149

cw/test_capture.sh
vogelpi: Thanks @vrozic for your suggestion, this is now changed.

cw/test_capture.sh
vogelpi: Thanks @vrozic and @wettermo for your input. I've included this as is and it seems to work okay. Just needed to add something to distinguish the result figures for keygen and modinv.

Regarding the points you've raised @vrozic :
1. This seems to be handled by the yaml file entirely and I believe it's fine as is
2. Good point, I believe testing `KEY` is sufficient. IIRC, this is what we really care about here.
3. Thanks, sounds good.
4. Thanks, I've added a TODO in the script and will update/merge that other PR in OpenTitan.

vrozic (approved):
Thanks @vogelpi for the update.
I've tested the latest changes  and the script works well.
All graphs look correct except for the sha3-fvsr-data (non-batch) where it seems that the wrong offset is used. This is a pre-existing problem.

In my view this is ready to merge.

vogelpi:
Thanks @vrozic , yes I've noted the thing with the sha3 non-batch capture as well.

PR #153, opened 2023-08-18
[ci] Add test for aes-general tvla

vrozic:
Adds a CI test to open AES project, perform TVLA general analysis with 10 steps, and compare the results with a checked-in file.


vogelpi (approved):
LGTM, thanks @vrozic .

PR #154, opened 2023-09-08
[tvla] minor clean-up changes 

vrozic:
This PR consists of two commits with minor clean-ups for tvla.py.

1. Moves all functions for computing t-tests into a separate file
2. Makes 'HAMMING_WEIGHT' the default model for AES leakage

### DIFF
--- a/cw/tvla.py
+++ b/cw/tvla.py
@@ -7,6 +7,7 @@ import inspect
 import logging as log
 import multiprocessing
 import os
+import sys
 from pathlib import Path
 from types import SimpleNamespace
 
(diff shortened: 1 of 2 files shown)


cw/tvla.py
```diff
@@ -17,7 +17,7 @@
 import yaml
 from chipwhisperer.analyzer import aes_funcs
 from joblib import Parallel, delayed
-from scipy.stats import ttest_ind_from_stats
+from util_tvla.ttest import ttest_hist_xy
```
vogelpi, line 20: I don't have a strong opinion about this but I think it would make sense to actually move this `ttest.py` out of the `cw` folder into the `util` folder at the top level of the repo. Because it's not CW specific. Also it doesn't need to go into a tvla-specific utility folder. WDYT?

vogelpi (approved):
Thanks @vrozic , LGTM.

I have a question regarding where to place the `ttest.py` file. We could move it to a different folder as this is not CW-specific.

cw/tvla.py
vrozic: I agree. I've moved 'ttest.py' to `util`

PR #156, opened 2023-09-15
Compute leakage general

vrozic:
At the moment we are computing AES leakage in a separate function `compute_leakage_aes()`, while leakage for TVLA general test is computed directly in `run_tvla()`.
This PR adds `compute_leakage_general()` function.
This change is necessary to set the stage for future changes - we plan to add more commands that will make calls to `compute_leakage_general()` and `compute_leakage_aes()`.


wettermo (approved):
LGTM! Makes the code a lot easier to extend in the future when having such functions to call. Good idea!

PR #157, opened 2023-09-18
[capture] Remove redundant key write

vrozic:
Writing key at the begining of `capture_kmac_random()` is redundant because `cw.capture_trace()` will send the key on every run.

### DIFF
--- a/cw/capture.py
+++ b/cw/capture.py
@@ -892,7 +892,6 @@ def capture_kmac_random(ot, ktp):
     """
     key, _ = ktp.next()
     tqdm.write(f'Using key: {binascii.b2a_hex(bytes(key))}')
-    ot.target.simpleserial_write('k', key)
     while True:
         _, text = ktp.next()
         ret = cw.capture_trace(ot.scope, ot.target, text, key, ack=False, as_int=True)


cw/capture.py
```diff
@@ -892,7 +892,6 @@ def capture_kmac_random(ot, ktp):
     """
     key, _ = ktp.next()
     tqdm.write(f'Using key: {binascii.b2a_hex(bytes(key))}')
-    ot.target.simpleserial_write('k', key)
     while True:
         _, text = ktp.next()
         ret = cw.capture_trace(ot.scope, ot.target, text, key, ack=False, as_int=True)
```
wettermo, line 897: In general, makes sense and LGTM!
But how does `cw.capture_trace` actually send key/text to the device? Also via `simpleserial_write()`? And how is this handled in `kmac_serial.c`?

wettermo (approved):

cw/capture.py
vrozic: That's exactly how it works. `capture_trace()` uses method `set_key()` under the hood, which uses `simpleserial_write('k', key)`.
On the `kmac_serial.c` side, I think that it was just executing the `'k'` command twice writing the same key value.
It wasn't slowing down capture by much, but it was still unnecessary.

PR #162, opened 2023-09-25
Waverunner: Add functions for load/store of setups to local file

johannheyszl:
incl. test file

### DIFF
--- a/cw/test_waverunner.py
+++ b/cw/test_waverunner.py
@@ -1,0 +1,25 @@
+#!/usr/bin/env python3
+# Copyright lowRISC contributors.
+# Licensed under the Apache License, Version 2.0, see LICENSE for details.
+# SPDX-License-Identifier: Apache-2.0
+
+"""Test WaveRunner class."""
+
+from datetime import datetime
+
+from waverunner import WaveRunner
+
+if __name__ == '__main__':
+    waverunner = WaveRunner("172.26.111.125")
+
+    # Save WaveRunner setup to timestamped file
+    now = datetime.now()
+    now_str = now.strftime("%Y-%m-%d_%H:%M:%S")
+    file_name_local = f"scope_config_{now_str}.lss"
+    waverunner.save_setup_to_local_file(file_name_local)
+
+    # Write setup to local file for later restore (no time-stamps)
+    file_name_local = "scope_config.lss"
+    waverunner.save_setup_to_local_file(file_name_local)
+    # Load setup from local file
+    waverunner.load_setup_from_local_file(file_name_local)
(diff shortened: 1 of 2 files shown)


cw/waverunner.py
```diff
@@ -92,7 +94,71 @@ def _ask(self, cmd):
...
+            "Unrecognized keyword.", 
+            "String error.", 
+            "GET embedded in another message.", 
+            "Arbitrary data block expected.", 
+            "Non-digit character in byte count field of arbitrary data block.", 
+            "EOI detected during definite length data block transfer.", 
+            "Extra bytes detected during definite length data block transfer"]
+        return_code = int((re.findall(r'\d+', self._ask("CMR?")))[0])
+        if return_code > 11:
+            return_msg = f"{return_code} Unkown error code"
```
vrozic, line 115: Looking at the table in the manual, the codes are numbered 1..7 and 10..13, while codes 8 and 9 are unused.

vrozic:
Thanks @johannheyszl 
This looks mostly good FWICT, but of course I cannot test it with an oscilloscope.
I only noticed that the error codes may be inconsistent with documentation (comment above)

CI failures are mostly due to `flake8` complaining about indentation, trailing white spaces and imported-but-unused modules.

cw/waverunner.py
johannheyszl: mega - thanks for catching this!

cw/waverunner.py
```diff
@@ -93,6 +95,74 @@ def _ask(self, cmd):
...
+                     "String error.",
+                     "GET embedded in another message.",
+                     "",
+                     "",
+                     "Arbitrary data block expected.",
+                     "Non-digit character in byte count field of arbitrary data block.",
+                     "EOI detected during definite length data block transfer.",
+                     "Extra bytes detected during definite length data block transfer"]
+        return_code = int((re.findall(r'\d+', self._ask("CMR?")))[0])
+        if return_code > 11:
```
vrozic, line 118:
```suggestion
        if return_code > 13 or return_code in [8, 9]:
```

cw/test_waverunner.py
```diff
@@ -0,0 +1,24 @@
+#!/usr/bin/env python3
+# Copyright lowRISC contributors.
+# Licensed under the Apache License, Version 2.0, see LICENSE for details.
+# SPDX-License-Identifier: Apache-2.0
+
+"""Test WaveRunner class."""
+
+from datetime import datetime
+from waverunner import WaveRunner
```
vrozic, line 9:
```suggestion
from datetime import datetime

from waverunner import WaveRunner
```

I'm afraid that the remaining lint error message is not very informative.
You can run `isort -c -w79 test_waverunner.py --diff` locally to see what isort is expecting.
At the moment it is only complaining about the missing empty line.

johannheyszl:
Thanks @vrozic for your help

vrozic (approved):

PR #166, opened 2023-10-03
[docker/podman] Fix user permissions

nasahlpa:
Before this commit, the container started as root with the following warnings:
```
Creating user 'ot' with UID=0, GID=0. 
groupadd: GID '0' already exists 
useradd: UID 0 is not unique
runuser: user ot does not exist
root@ot-sca:/repo#
```
With podman 4.3, the option --userns=keep-id when running the container forwards the UID & GID.
Hence, no manual permission management is needed anymore.
Now, the user inside the container matches the user starting it.


vogelpi (approved):
LGTM, many thanks @nasahlpa !

PR #172, opened 2023-10-06
[tvla] Enable filtering

vrozic:
At the moment trace filtering is enabled only for aes-specific tvla. This PR enables it for general analysis as well.


nasahlpa (approved):

vrozic:
Thanks @nasahlpa for the review. I will merge this now

PR #173, opened 2023-10-09
[cw] Update CW Python dependency

nasahlpa:
This PR updates the ChipWhisperer Python dependency to the latest available CW version.

For Husky, we need to make sure that the clkgen is locked after setting the clock.

### DIFF
--- a/cw/util/device.py
+++ b/cw/util/device.py
@@ -152,6 +152,10 @@ class OpenTitan(object):
         scope.io.tio2 = "serial_rx"
         scope.io.hs2 = "disabled"
 
+        # Make sure that clkgen_locked is true.
+        if husky:
+            scope.clock.clkgen_src = 'extclk'
+
         # TODO: Need to update error handling.
         if not husky:
             scope.clock.reset_adc()
--- a/python-requirements.txt
+++ b/python-requirements.txt
@@ -27,7 +27,7 @@ joblib
 # Development version of ChipWhisperer toolchain with latest features and
 # bug fixes - Needs to be installed in editable mode. We fix the version
 # for improved stability and manually update if necessary.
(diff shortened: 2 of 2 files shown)


nasahlpa:
Waiting until #PR [466](https://github.com/newaetech/chipwhisperer/pull/466) gets merged.

cw/util/device.py
```diff
@@ -152,6 +152,10 @@ def initialize_scope(self, scope_gain, num_samples, offset, pll_frequency):
         scope.io.tio2 = "serial_rx"
         scope.io.hs2 = "disabled"
 
+        # Make sure that clkgen_locked is true.
+        if husky:
+            scope.clock.clkgen_src = 'extclk'
+
```
vogelpi, line 158: Is this right? The comment seems not to be in line with the code. Also, we already set `scope.clock.clkgen_src = 'extclk'` above on Line 127.

vogelpi (approved):
Thanks for doing this @nasahlpa . LGTM with one question.

cw/util/device.py
nasahlpa: Without this code line, the PLL was not locked making the ADC unable to lock itself. According to [this](https://forum.newae.com/t/adc-freq-setting/3108/2), again setting `clkgen_src` fixes the issue, which also worked on my site.

johannheyszl: agree, odd, seems to be done twice now. does the if in L125 work properly?
I'd guess it evaluates false there even on husky?

johannheyszl: I think this is probably what prevented me from separating CW310 and Husky into two separate classes ... so I'd be interested :)

johannheyszl: thx, so L130 is in fact executed I understand.
tanks again - maybe will try split the two CW310 and Husky into two separate classes instead of OpenTitan in device.py later with this fix

johannheyszl (approved):

PR #174, opened 2023-10-09
[ci] experiment

vrozic:
extend CI with AES, KMAC, SHA3 and OTBN capture tests


nasahlpa (approved):
That looks good, thanks!

PR #175, opened 2023-10-10
[capture] Add drastically simplified capture script as template for post-si testing

johannheyszl:
Derived from capture.py, this PR adds a simplified script (with a yaml cfg file) which performs only the `aes-random` command from capture.py. It shall serve as starting point and template for script development for post-si development. It represents some of the desired changes from the [design doc](https://docs.google.com/document/d/1HjsIG-4aXvQZqbH-zv8tTlzKCQEE1RdwtXapCEydKFU/edit?usp=sharing). Will be used to replace scope with WaveRunner and target by post-si OT silicon later.

- device.py is slightly modified to include what was previously in capture.py
- trace_util.py is new to collect utility functions


vrozic:
I've just tested the script and it seems to work well.
Here is the TVLA result for 10k traces.

aes_t_test_round_0_byte_0

PR #178, opened 2023-10-11
[cw] Remove CW-Lite support

### LINKED ISSUE #67: Deprecate CW-Lite support
All of us now have access to the Husky scope and we almost exclusively use Husky instead of CW-Lite due to the better ADC resolution, higher sampling frequency and higher capture rate (larger on-chip buffer). Therefore we should consider to deprecate the CW-Lite support to clean up the repository.

nasahlpa:
As Husky has better capture capabilities, this PR removes the support for the ChipWhisperer Lite to close #67.


vrozic:
Thanks @nasahlpa for working on fixing this. The PR looks good to me.

Would it make sense to also edit the `Typical usage:` comment in cw_segmented.py?
Right now it uses variable `cw_lite`, I suggest to rename it to something like `scope` or `cw_scope`, just to reduce the possibility of confusion later on.

vrozic (approved):

nasahlpa:
> Thanks @nasahlpa for working on fixing this. The PR looks good to me.
> 
> Would it make sense to also edit the `Typical usage:` comment in cw_segmented.py? Right now it uses variable `cw_lite`, I suggest to rename it to something like `scope` or `cw_scope`, just to reduce the possibility of confusion later on.

Yes, you are right - thanks!
I changed this.

PR #179, opened 2023-10-11
[ci] Automatic comparison of traces

### LINKED ISSUE #176: [ci] Automatic comparison of traces
Currently, traces generated in the CI need to be manually inspected. Ideally, an automated comparison of generated traces to a golden model is conducted.

nasahlpa:
This PR adds a new CI job responsible for comparing captured AES traces to a reference golden trace set. This enables us to quickly detect issues with the setup.
- 100 AES traces on CW310 are captured
- These traces as well as the golden traces are imported
- The mean over each trace set is calculated
- The correlation factor of the resulting two traces is determined
- If the correlation factor is below a threshold (currently 0.8), the job fails
- The traces and the CW project are provided as an asset

This closes #176.


vrozic (approved):
Thanks @nasahlpa for working on this.

LGTM

PR #180, opened 2023-10-11
[capture] Added WaveRunner support to simple capture script and cfg file

johannheyszl:
Added WaveRunner support to `simple_capture_aes_sca.py`. Non-batch/sequence/segmented only for now. Both scopes tested.

Small extension in waverunner.py to support single trace capture.


nasahlpa (approved):
I don't have a Waverunner scope to test, but the code itself LGTM.

PR #182, opened 2023-10-13
[capture] Fix project export

nasahlpa:
As not all config files yet contain the export keyword, this PR adds code that first checks whether this config entry exists.


johannheyszl (approved):
helps :)

PR #183, opened 2023-10-13
[ci] Add TVLA AES check to CI

nasahlpa:
This PR runs TVLA for AES and publishes the TVLA figure as an artifact. Expected to fail as currently TVLA seems to be broken.


vrozic:
Test failed successfully. :)
Now that #184 is in, we should expect it to pass after rebase.

vrozic (approved):
Thanks @nasahlpa for working on this
LGTM

PR #185, opened 2023-10-13
[tvla] Make trace filtering optional.

vrozic:
Before starting the analysis, tvla filters out the noisy traces.
This PR makes trace-filtering optional via input argument. The default value is True.


nasahlpa (approved):
LGTM, thanks.

PR #192, opened 2023-10-16
[doc] Add documentation for using CI

jwnrt:
This repo's CI has a couple of quirks that I've tried to document here:

1. We're using a self-managed pool of FPGAs containing both a CW305 and CW310. Selecting the correct one requires matching a tag.
2. We don't run CI for pull requests from forks created by users outside of the lowRISC GitHub organisation. These must be manually approved before they run.

Please let me know if anything's unclear, if this should go elsewhere, or if there's something else you would like documenting.

### DIFF
--- a/doc/ci.md
+++ b/doc/ci.md
@@ -1,0 +1,45 @@
+# Continuous Integration (CI)
+
+This repository has access to a pool of FPGAs in Azure Pipelines for running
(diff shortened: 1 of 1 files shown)


vrozic (approved):
Thanks @jwnrt 
This is very helpful

PR #193, opened 2023-10-17
[ci] Add filter_traces to CI job

nasahlpa:
Add filter_traces to CI job cfg and increase number of trace captures to 5000 to get consistent TVLA results.


vrozic (approved):
LGTM

PR #195, opened 2023-10-18
[device] CW340 support

nasahlpa:
This PR adds experimental CW340 support with a minimal running capture script.


nasahlpa:
Hi @Blebowski,

Thanks for your interest in ot-sca!

Currently, we are not planning to support CW340 in ot-sca. The reason is that capturing power traces on CW340 is much more challenging (i.e. more noise) due to the larger FPGA and the larger design. Hence, when capturing traces, we use the smaller OpenTitan design for CW310 and even remove some IPs that we are not measuring to minimize the noise.

However, adding support for CW340 by yourself should be quite straightforward - this draft PR should give you an idea how this could work.

Blebowski:
Thanks @nasahlpa.

My thinking was precisely opposite. Since ultra-scale has better clocking, it would allow to emulate
e.g. per-IP clock gate enable as on the ASIC, and therefore not to cause noise from other IPs.

But I guess that Open Titan does not have per-peripheral manual gating(that needs to be functionally enabled
for peripheral to work) and relies only on inserted clock gates ?

PR #197, opened 2023-10-23
[Capture] Add fvsr key with random plaintext testing to simple script as option

johannheyszl:
Readability of simple capture script lower but wanted to add to have it in template.
Hints for improvement of structure / readability welcome.

Capture tested on Husky and WaveRunner.


nasahlpa (approved):
LGTM.
It is still short & readable.

However, I am not sure how we want to proceed in the future when we analyze different ciphers. Do we want to have a similar script for each cipher?

johannheyszl:
Thanks @nasahlpa.

Re future and different ciphers. If we are happy with the script at some point we could simply copy and have different copies per test class, i.e. cipher. I have put a lot in this one so we have a template to copy and generate simpler ones through removal of some stuff if needed. But let's discuss at some point - maybe there is a better way :)

Issue #177, opened 2023-10-10
[capture] Specify offset in clk cycles


vrozic:
At the moment, measurement offset in capture configuration files is specified in samples.
It probably makes more sense to use clk cycles of the target as a unit of measurement.

vogelpi:
@vrozic is working on this and we've briefly discussed this today. Even if for the time being the computation of the offsets and number of samples / window size happens in / close to the scope code this would be okay. We can refine this later.

Ideally we have a first working version for this soon. We need this to re-enable captures with latest bitstreams (use different target clock frequencies on the CW310). My goal is to work on this in the second half of this week.

vogelpi:
Update: we merged PR #203 which implements the corresponding changes in the CW-based setup. What's left to do is to align the WaveRunner implementation.

vogelpi:
I am thus un-assigning myself and @vrozic and instead assign @johannheyszl who might be able to take care of the WaveRunner part when returning to work.

johannheyszl:
has been addressed in #242

PR #203, opened 2023-10-25
[capture] Specify offset in target clk cycles

vrozic:
Specify offset in target clock cycles, rather than number of samples.
Partially addresses #177 
For the time being, I've added offset adjustment in the scope code (`cw_segmented.py` and `device.py`), because that's where we can read the sampling rate.
Perhaps there is an elegant way to do this computation without touching the scope code, but I don't see it at the moment.

### DIFF
--- a/cw/simple_capture_aes_sca.yaml
+++ b/cw/simple_capture_aes_sca.yaml
@@ -5,9 +5,10 @@ cwfpgahusky:
   baudrate: 115200
   output_len_bytes: 16
   pll_frequency: 100000000
+  target_clk_mult: 0.1
   num_segments: 20
-  num_samples: 1200
-  offset: -40
+  num_cycles: 60
+  offset_cycles: -2
   scope_gain: 38
 waverunner:
   waverunner_ip: 100.107.71.10
--- a/cw/util/device.py
+++ b/cw/util/device.py
@@ -11,6 +11,9 @@ import chipwhisperer as cw
 
 from util.spiflash import SpiProgrammer
 
+PLL_FREQUENCY_DEFAULT = 100e6
+SAMPLING_RATE_MAX = 200e6
+
 
 class RuntimePatchFPGAProgram:
     """Replaces the FPGAProgram method of an FPGA object with a function
@@ -44,8 +47,8 @@ class RuntimePatchFPGAProgram:
 
 class OpenTitan(object):
 
-    def __init__(self, bitstream, force_programming, firmware, pll_frequency,
-                 baudrate, scope_gain, num_samples, offset, output_len):
+    def __init__(self, bitstream, force_programming, firmware, pll_frequency, target_clk_mult,
+                 baudrate, scope_gain, num_cycles, offset_cycles, output_len):
 
         # Extract target board type from bitstream name.
         m = re.search('cw305|cw310', bitstream)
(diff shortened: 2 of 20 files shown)


vogelpi:
Thanks @vrozic for initiating this. I've taken your work as a basis and continued the effort. I've now pushed back the result to this PR.

With this PR, the code now takes the PLL frequency of the target ( = ADC base frequency), multiplies that by an integer to get as close to 200 MHz (max sampling frequency of Husky) and then computes the number of samples to take as well as the horizontal offset based on numbers specified in target clocks. The nice thing about this is that we can very easily change the base frequency to get more oversampling without changing the offsets / window length.

What is missing from this PR is
1. the adjustment of the simplified capture script - @nasahlpa offered to take care of this
2. the corresponding changes to make this work with WaveRunner - @johannheyszl offered his help but he is OoO this week.

I think we should try to merge this asap to unblock other work.

vogelpi:
Update: also the simple capture script now works with this.

nasahlpa (approved):
Thanks @vogelpi & @vrozic - LGTM!
Test with simple_script and capture.py works for me.

vogelpi:
Thanks for the review @nasahlpa , let's merge this :-)

cw/util/device.py
```diff
@@ -60,15 +63,35 @@ def __init__(self, bitstream, force_programming, firmware, pll_frequency,
...
+        self.adc_mul = int(SAMPLING_RATE_MAX // pll_frequency)
+        self.sampling_rate = self.clkgen_freq * self.adc_mul
+
+        # The target runs on the PLL clock but uses internal clock dividers and
+        # multiplier to produce the clock of the target block.
+        self.target_freq = pll_frequency * target_clk_mult
+
+        # The scope is configured in terms of samples. For Husky, the number of
+        # samples must be divisble by 3 for batch captures.
+        # TODO: For WaveRunner, we need to read the configured sampling rate here
```
johannheyszl, line 82: nit: for Waverunner we do this outside of this class, in the capture script when we configure the waverunner scope

cw/util/device.py
```diff
@@ -60,15 +63,35 @@ def __init__(self, bitstream, force_programming, firmware, pll_frequency,
...
+        # In our setup, Husky operates on the PLL frequency of the target and
+        # multiplies that by an integer number to obtain the sampling rate.
+        # Note that the sampling rate must be at most 200 MHz.
+        self.clkgen_freq = pll_frequency
+        self.adc_mul = int(SAMPLING_RATE_MAX // pll_frequency)
+        self.sampling_rate = self.clkgen_freq * self.adc_mul
+
+        # The target runs on the PLL clock but uses internal clock dividers and
+        # multiplier to produce the clock of the target block.
+        self.target_freq = pll_frequency * target_clk_mult
```
johannheyszl, line 78: nit: is this correct? pls double check. I get a clock of 20 MHz on OT with pll_frequency: 100000000 and target_clk_mult: 0.1

cw/simple_capture_aes_sca.yaml
```diff
@@ -5,9 +5,10 @@ cwfpgahusky:
   baudrate: 115200
   output_len_bytes: 16
   pll_frequency: 100000000
+  target_clk_mult: 0.1
```
johannheyszl, line 8: with this config, it captures 120 clock cycles (can be counted in the plot window) since the target_clk_mult does not match the actual clk divider in the FPGA. The clk frequency is actually 20 MHz instead of 10 Mhz as suggested here.

IMO correct would be to set `target_clk_mult: 0.2` and `num_cycles: 120` and `offset_cycles: -4`

vogelpi: I am 100% sure it's capturing 60 target clock cycles with these settings (as before!). The AES encryption matches pretty nicely into the capture window. AES takes exactly 56 clock cycles. If we would capture 120 clock cycles, the AES trace would only span half the capture window / the other half would be "empty".

Can you please elaborate on why you think to observe 20 MHz target clock frequency? I would like to understand why we have different observations here. Maybe you want to upload the captured wave plot?

johannheyszl: Thanks @vogelpi, adding from our offline chat: 
- We have a 50% duty cycle clock on the target. 
- Also it seems the inactive clock edge creates a similarly significant peak as the active clock edge. 
- Hence we need to expect two significant peaks per clock period and the original settings check out :)
image

cw/util/device.py
johannheyszl: thanks!

PR #204, opened 2023-10-25
Update binaries for OTBN horizontal tests

vrozic:
This PR adds several changes:
-Adds a binary for OTBN ECDSA384 test.
-Updates a binary for OTBN ECDSA256 test - the old binary was not compatible with the bitstream
(both binaries were generated from [https://github.com/vrozic/opentitan/tree/binaries-ecc384](https://github.com/vrozic/opentitan/tree/binaries-ecc384)  )
-Re-names some files for consistency
-Adds OTBN horizontal tests to `test_capture.py`

New binaries can be tested by running:
`./test_capture.py`
or
`./capture.py --cfg-file capture_ecdsa256_cw310.yaml  capture ecdsa-simple --num-traces 10`
`./capture.py --cfg-file capture_ecdsa256_cw310.yaml  capture ecdsa-stream --num-traces 10`
`./capture.py --cfg-file capture_ecdsa384_cw310.yaml  capture ecdsa-simple --num-traces 10`
`./capture.py --cfg-file capture_ecdsa384_cw310.yaml  capture ecdsa-stream --num-traces 10`


vogelpi (approved):
LGTM, I've also tested this on the FPGA and it seems to work all fine. Thanks @vrozic taking care of this!

I think we should also add these commands to CI. But this can be done as a follow-up.

PR #205, opened 2023-10-25
[cw] Introduce new trace database format

### LINKED ISSUE #202: [capture] Trace storage: Future-proof solution
We currently hold all traces in RAM after capture / during analysis and write to CW projects, i.e. numpy files. Larger trace amounts require sequential iterations through traces to reduce RAM consumption.

nasahlpa:
As [discussed](https://docs.google.com/document/d/1sxLxlU3x4cLbHQCDNusWYLsQ9gkUJq8kTgY4X9QKMcE/edit), this PR introduces a new database format based on SQLite. A helper class provides an interface to the DB. Traces are held in memory until a user defined threshold is reached, then the traces are flushed into storage.
Closes #202.

- `cw/capture.py` is using the old CW project file
- `cw/simple_capture_aes_sca.py` is using the new trace library
- `cw/tvla.py` can handle both file formats


cw/util/trace_library.py
```diff
@@ -0,0 +1,110 @@
...
+            self.commit_traces()
+    
+    def get_traces(self, start=0, end=0):
+        """ Write traces into database.
+
+        The trace first is written into memory. When reaching the trace
+        memory threshold, the traces are flushed into the database storage.
+
+        Args:
+            trace: The trace to write.
```
johannheyszl, line 93: nit: Args

cw/util/trace_library.py
```diff
@@ -0,0 +1,110 @@
...
+        """
+        self.trace_mem.append(trace)
+        self.trace_mem_ctr += 1
+        if (self.trace_mem_ctr >= self.trace_mem_thr):
+            self.commit_traces()
+    
+    def get_traces(self, start=0, end=0):
```
johannheyszl, line 86: later, if we start using x and y, we would probably want to have them as optional args here

johannheyszl (commented):
This is great @nasahlpa !

One small thing: Should we make the choice b/w Chipwhisperer and new Trace DB configurable in the simple capture script for now until e.g. TVLA can read the new format?

nasahlpa:
> This is great @nasahlpa !
> 
> One small thing: Should we make the choice b/w Chipwhisperer and new Trace DB configurable in the simple capture script for now until e.g. TVLA can read the new format?

Good point, I will adapt the capture script accordingly.
I guess it also would make sense to store additional metadata into the DB.  Datetime & samplerate would be nice, what else?

johannheyszl:
Yeah could be useful and does not hurt. The script already saves the entire scope config to a separate file though. Maybe something to extend later when we know what we need :)

cw/util/trace_library.py
nasahlpa: They are already optional. When not providing start and end, all available traces are returned.

nasahlpa:
> Yeah could be useful and does not hurt. The script already saves the entire scope config to a separate file though. Maybe something to extend later when we know what we need :)

So we want to keep the CW project file? We could get rid of this file by storing the information directly into the DB.

cw/util/trace_library.py
```diff
@@ -0,0 +1,194 @@
...
+
+
+class TraceLibrary:
+    """ Class for writing/reading traces to a database.
+
+    Either a new database is created or traces are appended to the existing
+    one. Traces are first written into memory and the flushed into the
+    database after reaching a trace memory threshold.
+    """
+    def __init__(self, db_name, trace_threshold):
```
rswarbrick, line 55: Do we ever expect `TraceLibrary` to read an existing file? If not, maybe it would make sense to move the unlink stuff from `simple_capture_aes_sca.py` to live here. It might make the library a bit easier to use.

cw/util/trace_library.py
```diff
@@ -0,0 +1,194 @@
...
+        self.engine = db.create_engine("sqlite:///" + db_name + ".db")
+        db_utils.create_database(self.engine.url)
+        self.Session = sessionmaker(self.engine)
+        self.session = self.Session()
```
rswarbrick, line 59: Is there a need for having both capitalisations? If not, maybe it's easier to pick one?

cw/util/trace_library.py
```diff
@@ -0,0 +1,194 @@
...
+        self.metadata = db.MetaData()
+        self.TracesDB = db.Table(
```
rswarbrick, line 61: Nit: Would a name like "traces_table" make more sense? (It's a table inside a database, rather than the database itself)

cw/util/trace_library.py
```diff
@@ -0,0 +1,194 @@
...
+        self.metadata.create_all(self.engine)
+        self.trace_mem = []
+        self.trace_mem_ctr = 0
+        self.trace_mem_thr = trace_threshold
+        self.settingsDict = {}
+
+    def commit_traces(self):
+        """ Writes traces from memory into database.
+        """
+        if len(self.trace_mem):
```
rswarbrick, line 95: This might be clearer as just `if self.trace_mem` (since an empty list is "false" for this sort of test)

cw/util/trace_library.py
```diff
@@ -0,0 +1,194 @@
...
+            query = db.insert(self.TracesDB)
+            traces = []
+            for trace in self.trace_mem:
+                tr = asdict(trace)
+                del tr["trace_id"]
+                traces.append(tr)
```
rswarbrick, line 101: I wonder if it would be simpler to do this conversion inside `write_trace`. If I understand correctly, we only use `trace_mem` as an input for this function, so converting the traces to dictionaries and dropping the unused ids as the traces get appended might make things slightly simpler.

cw/util/trace_library.py
```diff
@@ -0,0 +1,194 @@
...
+        """ Write traces into database.
+
+        The trace first is written into memory. When reaching the trace
+        memory threshold, the traces are flushed into the database storage.
+
+        Args:
+            trace: The trace to write.
+        """
+        self.trace_mem.append(trace)
+        self.trace_mem_ctr += 1
```
rswarbrick, line 117: Do we need this counter? I think it's always equal to `len(self.trace_mem)`?

rswarbrick (commented):
This looks sensible. I started making some comments, but I think I'm racing against you! Sorry! I'll get out of the way until the PR is ready :-)

cw/util/trace_library.py
```diff
@@ -0,0 +1,221 @@
...
+            The bytearray waves from the database.
+        """
+        self.commit_traces()
+        traces = self.get_traces()
+        waves = []
+        for trace in traces:
+            waves.append(trace.wave)
+        return waves
+
+    def get_waves_int(self):
```
vogelpi, line 162: More general question (putting it here because I couldn't find out where the data format is defined): What is the storage format? Is it doubles or floats? In my view it should either be int16 or when using 8-bit scopes int8. Everything else is probably very wasteful in terms of memory, disk space and compute (when converting on the fly upon reading). To be honest, I think int8 could even be skipped in favor of just int16 to simplify things.

vogelpi (commented):
Thanks for working on this @nasahlpa . I think this is generally going in the right direction. I've left some general comments but @rswarbrick can probably give more valuable feedback on the actual Python code.

cw/util/trace_library.py
nasahlpa: Yes, we sometimes want to load an existing table. It might be an option to pass over an argument to the class indicating whether we want to overwrite the database or not.

cw/util/trace_library.py
```diff
@@ -0,0 +1,194 @@
...
+class Metadata:
+    """ Metadata for capture.
+
+    Stores information about the capture of the traces.
+
+    """
+    config: str
+    datetime: datetime.datetime
+    bitstream: str
+    binary: str
```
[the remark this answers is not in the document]
nasahlpa: Hm usually it is not a good practice to store large binary blobs (such as the actual firmware binary) into a database. However, as we do not really care about the database size, it might be an option. Do we want to discuss this in the next SCA/FI meeting?
Commit of the ot-sca repo is a good idea, I'll add it.

cw/util/trace_library.py
```diff
@@ -0,0 +1,194 @@
...
+            "traces",
+            self.metadata,
+            db.Column("trace_id", db.Integer, primary_key=True,
+                      autoincrement=True),
+            db.Column("wave", db.LargeBinary),
+            db.Column("plaintext", db.LargeBinary),
+            db.Column("ciphertext", db.LargeBinary),
+            db.Column("key", db.LargeBinary),
+            db.Column("x_pos", db.Integer),
+            db.Column("y_pos", db.Integer),
```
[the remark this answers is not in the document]
nasahlpa: I'll add a comment. These are the x/y table position for potential EM experiments.

cw/util/trace_library.py
nasahlpa: The idea is to store traces in memory (=`trace_mem`) to make things faster. When writing traces to the storage (database), we do not really care about speed. Therefore, the conversion is currently done in the `commit_traces `function. However, in reality, I do not think that there is a big performance impact. But I keep it as it is for now.

cw/util/trace_library.py
```diff
@@ -0,0 +1,194 @@
...
+            for trace in self.trace_mem:
+                tr = asdict(trace)
+                del tr["trace_id"]
+                traces.append(tr)
+            self.session.execute(query, traces)
+            self.session.commit()
+            self.trace_mem_ctr = 0
+            self.trace_mem = []
+
+    def write_trace(self, trace):
```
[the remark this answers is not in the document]
nasahlpa: Good point, I renamed the functions.

nasahlpa:
Thanks for the comments, @vogelpi & @rswarbrick - they are now addressed. The new trace database is now also tested by the CI.

cw/util/trace_library.py
rswarbrick: Ahah! I hadn't twigged that we might not actually do the conversion every time. I'm convinced :-)

cw/util/trace_library.py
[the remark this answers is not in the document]
vogelpi: Good idea to discuss this today in the SCA meeting.

cw/util/trace_library.py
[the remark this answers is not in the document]
vogelpi: Ah, I see. Thanks for explaining.

cw/util/trace_library.py
vogelpi: Thanks for explaining, I wasn't aware that we don't store numpy files. Follow-up question: is everything in one file? What happens if you you've stored 100k traces to disk already and you want to append the next 100k traces, will it store all 200k traces again (because it's just one file)?

nasahlpa: The data is stored in a single SQL database file.
When using `write_to_buffer`, traces are stored in memory. With `flush_to_disk`, this memory buffer is committed into the database file. If there are already 100k traces stored and you want to flush another 100k traces from memory into the database file,  `flush_to_disk` just appends these 100k traces to the existing 100k.

vogelpi: This is perfect, thanks for explaining!

nasahlpa: Based on our discussion, I added fields for storing the binary and the bitstream.

cw/util/trace_library.py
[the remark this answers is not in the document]
nasahlpa: Based on our discussion, I added fields for storing the binary and the bitstream.

ci/cfg/ci_simple_capture_aes_sca.yaml
```diff
@@ -0,0 +1,40 @@
+cwfpgahusky:
```
vogelpi, line 1: Can you please create an issue to check whether we can unify the ci configs with the default configs currently living under cw/? I think it would save us quite some maintenance work by not having to continuously adjust two configs files.

ci/cfg/ci_tvla_cfg_aes_specific_byte0_rnd0.yaml
```diff
@@ -15,3 +15,5 @@ plot_figures: true
 general_test: false
 mode: aes
 filter_traces: true
+trace_threshold: 1000
+trace_db: ot_trace_library
```
vogelpi, line 19: Nit: missing new line at end of file.

cw/simple_capture_aes_sca.py
```diff
@@ -212,17 +230,26 @@ def abort_handler_during_loop(this_project, sig, frame):
...
+                    trace = Trace(wave=waves[i, :].tobytes(),
+                                  plaintext=text,
+                                  ciphertext=ciphertext,
+                                  key=key)
+                    project.write_to_buffer(trace)
+                else:
+                    # CW Trace
+                    trace = cw.Trace(waves[i, :], text, ciphertext, key)
+                    # Append trace Library trace to database
+                    # FIXME Also use uint16 as dtype for 8 bit WaveRunner for
```
vogelpi, line 245: Please change the FIXME to a TODO because thats what we use in the OpenTitan project.

cw/simple_capture_aes_sca.py
```diff
@@ -238,16 +265,37 @@ def abort_handler_during_loop(this_project, sig, frame):
...
                                cfg["capture"]["trace_image_filename"], add_mean_stddev=True)
         print(f'Created plot with {cfg["capture"]["plot_traces"]} traces: '
               f'{Path(cfg["capture"]["trace_image_filename"]).resolve()}')
 
-    # Save metadata and entire configuration cfg to project file ---------------
-    project.settingsDict['datetime'] = datetime.now().strftime("%m/%d/%Y, %H:%M:%S")
-    project.settingsDict['cfg'] = cfg
     if HUSKY:
         sample_rate = int(round(cwfpgahusky.scope.clock.adc_freq, -6))
+        sample_offset = cwfpgahusky.scope.adc.offset
```
vogelpi, line 281: IIRC `adc.offset` is just the positive horizontal offset. If you want to record earlier (i.e. negative horizontal offset) this is stored under `adc.pre_samples` or somenthing. Alternatively, you can use the `offset_samples` variable from the OpenTitan device class.

cw/simple_capture_aes_sca.py
```diff
@@ -238,16 +265,37 @@ def abort_handler_during_loop(this_project, sig, frame):
...
+    if WAVERUNNER:
+        sample_rate = cfg["waverunner"]["num_samples"]
+        sample_offset = cfg["waverunner"]["num_samples"]
```
vogelpi, line 284: This looks wrong...

cw/util/trace_library.py
```diff
@@ -0,0 +1,231 @@
...
+        waves = []
+        for trace in traces:
+            waves.append(trace.wave)
+        return waves
+
+    def get_waves_int(self):
+        """ Get all waves from the database in the trace array format.
+
+        Returns:
+            The int waves from the database.
```
vogelpi, line 176: Nit: it will return the data type defined in `wave_datatype`. So, it could also be float or double I guess. Maybe rename the functions to `get_waves()` and `get_waves_bytearray()`. In practice we will always use `get_waves()` and specify the datatype.

vogelpi (approved):
LGTM modulo some minor nits.

@vrozic would you mind reviewing this as well? In particular the TVLA changes might benefit from your input. Related to that, are we still running TVLA on both formats or only the new format now? In my view, we should test both formats at least for a while. Because TVLA is the most relevant analysis method right now.

nasahlpa:
> LGTM modulo some minor nits.
> 
> @vrozic would you mind reviewing this as well? In particular the TVLA changes might benefit from your input. Related to that, are we still running TVLA on both formats or only the new format now? In my view, we should test both formats at least for a while. Because TVLA is the most relevant analysis method right now.

Thanks @vogelpi . TVLA works with the old database format (CW) as well as with the new one. A new config entry ("trace_db") allows the user to specify which to use, e.g., `trace_db: ot_trace_library` for the new format or `trace_db: cw` for the old one.

ci/cfg/ci_simple_capture_aes_sca.yaml
nasahlpa: Done: #209.

cw/util/trace_library.py
```diff
@@ -0,0 +1,231 @@
...
+                del tr["trace_id"]
+                traces.append(tr)
+            self.session.execute(query, traces)
+            self.session.commit()
+            self.trace_mem = []
+
+    def write_to_buffer(self, trace):
+        """ Write traces into memory or into storage.
+
+        The trace first is written into memory. When reaching the trace
```
rswarbrick, line 121: Grammar nit: I think we probably want "The trace is first written ...". Also, the following sentence sounds like the traces are reaching the threshold. Maybe something like "When the length of the buffer reaches the memory threshold..."?

cw/util/trace_library.py
```diff
@@ -0,0 +1,231 @@
...
+        memory threshold, the traces are flushed into the database storage.
+
+        Args:
+            trace: The trace to write.
+        """
+        self.trace_mem.append(trace)
+        if (len(self.trace_mem) >= self.trace_mem_thr):
```
rswarbrick, line 128: Nit: Stray parens around the test here.

cw/util/trace_library.py
```diff
@@ -0,0 +1,231 @@
...
+        Args:
+            start: fetch traces from trace index start
+            end: to trace index end. If no start and end is provided, all
+                 all traces are returned.
+        Returns:
+            The traces from the database.
+        """
+        self.flush_to_disk()
+
+        if end == 0:
```
rswarbrick, line 145: I think this assumes that `start` and `end` are either both zero (so pick everything!) or give a sensible range. If we want to keep that assumption, maybe it makes more sense to pass something like `id_bounds: Optional[List[int, int]] = None` (assuming I've remembered the type syntax correctly :-D)

cw/util/trace_library.py
```diff
@@ -0,0 +1,231 @@
...
+            query = db.select(self.traces_table)
+        else:
+            query = db.select(self.traces_table).where(
+                (self.traces_table.c.trace_id >= start) &
+                (self.traces_table.c.trace_id <= end))
+
+        ResultProxy = self.session.execute(query)
```
rswarbrick, line 152: The capitalisation in the name is a bit shouty :-) But I don't think we actually need a name at all!
```python
        for trace in self.session.execute(query).fetchall():
...
```

And I think we can do better than that too :-D (untested)
```python
        return [Trace(**trace._mapping)
                for trace in self.session.execute(query).fetchall()]
```

cw/util/trace_library.py
```diff
@@ -0,0 +1,231 @@
...
+
+    def get_waves_bytearray(self):
+        """ Get all waves from the database.
+
+        Returns:
+            The bytearray waves from the database.
+        """
+        self.flush_to_disk()
+        traces = self.get_traces()
+        waves = []
```
rswarbrick, line 167: Again, I think this might be shorter with a list comprehension:
```python3
        return [trace.wave for trace in self.get_traces()]
```

Similarly in `get_waves` below. I *think* the whole lot could be this:
```python3
    def get_waves_bytearray(self):
        """ Get all waves from the database.

        Returns:
            The bytearray waves from the database.
        """
        return [trace.wave for trace in self.get_traces()]

    def get_waves(self):
        """ Get all waves from the database in the trace array format.

        Returns:
            The waves from the database in the type wave_datatype.
        """
        return [np.frombuffer(b, self.wave_datatype)
                for b in self.get_waves_bytearray()]
```

cw/util/trace_library.py
```diff
@@ -0,0 +1,231 @@
...
+            traces.append(Trace(**trace._mapping))
+        return traces
+
+    def get_waves_bytearray(self):
+        """ Get all waves from the database.
+
+        Returns:
+            The bytearray waves from the database.
+        """
+        self.flush_to_disk()
```
rswarbrick, line 165: I think this is the first thing in `get_traces`, so I suspect we don't need it here too.

cw/util/trace_library.py
```diff
@@ -0,0 +1,231 @@
...
+
+        Returns:
+            The int plaintexts from the database.
+        """
+        self.flush_to_disk()
+        traces = self.get_traces(start, end)
+        ptx = []
+        for trace in traces:
+            ptx.append(np.frombuffer(trace.plaintext, np.uint8))
+        return ptx
```
rswarbrick, line 197: I think this could equivalently be:
```python
        return [np.frombuffer(trace.plaintext, np.uint8)
                for trace in self.get_traces(start, end)]
```

Similarly for `get_keys_int`.

rswarbrick (commented):
Looks sensible. A couple of proposals about making the Python a bit simpler.

cw/simple_capture_aes_sca.py
```diff
@@ -19,21 +19,33 @@
...
 
 if __name__ == '__main__':
+    now = datetime.now()
+    cfg_file = sys.argv[1]
     # Load configuration from file
-    with open('simple_capture_aes_sca.yaml') as f:
+    with open(cfg_file) as f:
         cfg = yaml.load(f, Loader=yaml.FullLoader)
+    # Determine trace library
+    CWLib = True
```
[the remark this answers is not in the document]
nasahlpa: This makes sense, I am working on this in a separate PR.

cw/util/trace_library.py
```diff
@@ -0,0 +1,231 @@
...
+            db.Column("scope_gain", db.Float),
+            db.Column("trigger_level", db.Float),
+            db.Column("time_div", db.Float),
+            db.Column("ot_sca_commit", db.String),
+            db.Column("notes", db.String)
+        )
+        self.metadata.create_all(self.engine)
+        self.trace_mem = []
+        self.trace_mem_thr = trace_threshold
+        self.settingsDict = {}
```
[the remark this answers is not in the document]
nasahlpa: Thanks for the comment. Yes this dict is actually not used, I'll remove it. In a new PR, I create a project/trace class allowing the user to either use CW or the new trace format with the same interface.

vrozic (approved):
Thanks @nasahlpa for working on this. Well done!

The TVLA changes look good to me.
I've extensively tested the script and I can confirm that it works well for both old and new data formats.

nasahlpa:
Thanks for the reviews, merging this now into master.

cw/simple_capture_aes_sca.py
```diff
@@ -238,16 +265,37 @@ def abort_handler_during_loop(this_project, sig, frame):
 
     # Create and show test plot ------------------------------------------------
     # Use this plot to check for clipping and adjust gain appropriately
+    if CWLib:
+        proj_waves = project.waves
+    else:
+        proj_waves = project.get_waves()
```
johannheyszl, line 271: nit: could we pls load just #plot_traces here? :)

Super nice the whole thing!

Issue #201, opened 2023-10-23
[capture] Implement all needed capture.py-provided commands using simple capture script


johannheyszl:
When the simple capture script is mature enough, use as template to implement remaining capture commands.

nasahlpa:
I guess we can close this issue as we have separate issues for the ciphers missing the simple script implementation.

PR #207, opened 2023-10-30
Restructure repository

### LINKED ISSUE #14: Restructure the repo directories

nasahlpa:
This PR consists of two commits:

The first commit restructures the repository, as [discussed](https://docs.google.com/document/d/1HjsIG-4aXvQZqbH-zv8tTlzKCQEE1RdwtXapCEydKFU/edit#heading=h.hmqgofcm7uu9).
```
tree .
.
├── analysis (models, pre/post processing libraries)
│   ├── configs
├── attacks (mountable attacks)
│   ├── configs
├── capture (data capture)
│   ├── devices
│      └── chipwhisperer
│      └── waverunner
├── obj (access to precompiled bitstreams, etc, TBD)
├── target (target boards)
├── tests (integration tests, and end-to-end testing)
└── utils (utility libraries, e.g. cw to trs conversion scripts)
```
New classes for target, scope, and project are introduced.
Closes #14.

The second commits updates the simple capture script from #175 to use the new directory structure as well as the new trace database (#205).

The idea is to split up the  `cw/capture.py` into multiple simple capture scripts stored into `attacks/` (c.f., #201).
Hence, `cw/capture.py` is not working anymore.


vrozic:
My 2 cents: I would definitely go ahead and remove the `jupyter` directory. I don't think we've ever used that code for our analysis.

vogelpi (commented):
Thanks @nasahlpa for working on this. I've reviewed the scope and target code. Will also look at the capture script once you're done with that.

nasahlpa:
Thanks for the comments, @vogelpi! The adapted simple script will be a new PR. Please merge this PR after #205.

nasahlpa:
@johannheyszl and @vogelpi what do you think of _(i)_ the new repository structure as well as _(ii)_ the adapted simple AES capture script?

capture/capture_aes.py
```diff
@@ -0,0 +1,347 @@
...
+    """ Print plot of traces.
+
+    Printing the plot helps to adjust the scope gain and check for clipping.
+
+    Args:
+        project: The project containing the traces.
+        config: The capture configuration.
+    """
+    if config["capture"]["show_plot"]:
+        plot.save_plot_to_file(project.get_waves(),
```
johannheyszl, line 277: are we loading all waves to RAM again here? :)

capture/scopes/chipwhisperer/husky.py
```diff
@@ -0,0 +1,106 @@
...
+        # In our setup, Husky operates on the PLL frequency of the target and
+        # multiplies that by an integer number to obtain the sampling rate.
+        # Note that the sampling rate must be at most 200 MHz.
+        self.clkgen_freq = pll_frequency
+        self.adc_mul = int(SAMPLING_RATE_MAX // pll_frequency)
+        self.sampling_rate = self.clkgen_freq * self.adc_mul
+
+        # The target runs on the PLL clock but uses internal clock dividers and
+        # multiplier to produce the clock of the target block.
+        self.target_freq = pll_frequency * target_clk_mult
```
johannheyszl, line 28: So we don't forget to checkt: In the previous form there was an issue here where the actual target freq. seemed different from what is computed here

capture/project_library/project.py
```diff
@@ -0,0 +1,146 @@
...
+    def __init__(self, project_cfg: ProjectConfig) -> None:
+        self.project_cfg = project_cfg
+        self.project = None
+
+    def create_project(self):
+        """ Create project.
+
+        Create project using the provided path.
+        """
+        project = None
```
johannheyszl, line 45: nit: directly assign to self.project below and remove this line?

capture/project_library/project.py
```diff
@@ -0,0 +1,146 @@
...
+        if self.project_cfg.type == "cw":
+            project = cw.create_project(self.project_cfg.path,
+                                        overwrite=self.project_cfg.overwrite)
+        elif self.project_cfg.type == "ot_trace_library":
```
johannheyszl, line 49: nit: do we want an error msg if no valid selected, or will this just explode bc we then need to check all other cfg params too? :)

capture/capture_aes.py
nasahlpa: Good point, thanks for the comment. I've changed the code - now only `num_traces `are loaded into memory.

capture/scopes/chipwhisperer/husky.py
```diff
@@ -0,0 +1,106 @@
...
+    def __init__(self, scope_gain, num_cycles, num_segments, offset_cycles,
+                 pll_frequency, target_clk_mult):
+        self.scope = None
+        self.scope_gain = scope_gain
+        # In our setup, Husky operates on the PLL frequency of the target and
+        # multiplies that by an integer number to obtain the sampling rate.
+        # Note that the sampling rate must be at most 200 MHz.
+        self.clkgen_freq = pll_frequency
+        self.adc_mul = int(SAMPLING_RATE_MAX // pll_frequency)
+        self.sampling_rate = self.clkgen_freq * self.adc_mul
```
johannheyszl, line 24: :) i think the computed sampling rate will always equal SAMPLING_RATE_MAX, right?

capture/project_library/project.py
nasahlpa: Added an error msg. Config path is sanitized by the argument parser, but the other configs are currently not checked.

capture/scopes/chipwhisperer/husky.py
johannheyszl: in other words: sampling rate here is fixed at max but seems to be computed from inputs. There is no cfg param atm to adjust sampling rate. it will always be max.

johannheyszl: We could have the sampling rate as a cfg param and compute adc_mul from it similar to above.

capture/scopes/chipwhisperer/husky.py
johannheyszl: the root issue was that the target_clk_mult does not match what is happening on FPGA, see https://github.com/lowRISC/ot-sca/pull/203/files#r1385486125

capture/configs/aes_sca.yaml
```diff
@@ -0,0 +1,45 @@
...
+  target_type: cw310
+  fpga_bitstream: "../objs/lowrisc_systems_chip_earlgrey_cw310_0.1.bit"
+  force_program_bitstream: False
+  fw_bin: "../objs/aes_serial_fpga_cw310.bin"
+  pll_frequency: 100000000
+  baudrate: 115200
+  output_len_bytes: 16
+husky:
+  pll_frequency: 100000000
+  target_clk_mult: 0.1
```
johannheyszl, line 11: re `target_clk_mult`: here it is a param for the scope. we dont use it to adjust the target. arguably it could make sense to have a param for the CW target FPGA to adjust clock. but we need to know the target clk in scope too to compute samples from cycles.

capture/scopes/chipwhisperer/husky.py
nasahlpa: I agree, from the usability side, it would make sense for me that the user needs to specify `pll_frequency`, `target_freq` (user should now at which freq. OT is running), and desired `sampling_rate`. So the code would look like this:

```python
        self.scope = None
        self.scope_gain = scope_gain
        # In our setup, Husky operates on the PLL frequency of the target and
        # multiplies that by an integer number to obtain the sampling rate.
        # Note that the sampling rate must be at most 200 MHz.
        self.clkgen_freq = pll_frequency
        self.sampling_rate =  sampling_rate
        self.adc_mul = self.sampling_rate / self.clkgen_freq

        # The target runs on the PLL clock but uses internal clock dividers and
        # multiplier to produce the clock of the target block.
        self.target_freq = target_freq

        # The scope is configured in terms of samples. For Husky, the number of
        # samples must be divisble by 3 for batch captures.
        sampling_target_ratio = self.sampling_rate / self.target_freq
        self.offset_samples = int(offset_cycles * sampling_target_ratio)
        self.num_samples = int(num_cycles * sampling_target_ratio)
        if self.num_samples % 3:
            self.num_samples = self.num_samples + 3 - (self.num_samples % 3)

        self.num_segments = num_segments
        self.scope = None
```

capture/scopes/chipwhisperer/husky.py
nasahlpa: As discussed previously, we need to check with @vogelpi which `target_clk_mult` was used and where to find this parameter.

capture/scopes/chipwhisperer/husky.py
johannheyszl: i think this would be great, and prevent that we would need to hand pll_freq to a waverunner scope as well

johannheyszl: thx - understand that there are limitations here. Maybe making those limitations explicit as comments along with params which have self-explanatory names is still easier? :)

nasahlpa: We could sanity check the user input to avoid that `adc_mul `isn't an integer.

nasahlpa:
As discussed in yesterday's SCA/FI meeting, this PR is ready to merge. Waiting for final approval.

capture/README.md
```diff
@@ -0,0 +1,44 @@
...
+
+## Quick Usage
+
+Follow the steps described in the [`getting_started`](../doc/getting_started.md)
+documentation to setup the environment, the target, and the scope.
+
+When using OpenTitan on the CW310 and measuring traces using Husky, AES traces
+for a random key tests can be captured with the following command:
+```console
+mkdir -p projects/
```
johannheyszl, line 14: nit: i think you dont even need to mkdir

capture/configs/simple_capture_aes_sca.yaml
johannheyszl, line 1: nit: I thought this file is obsolete?

johannheyszl (approved):
thanks again @nasahlpa !

LGTM and let's wait @vogelpi 's approval

We have opens re cfg params for clocking which can be addressed in a later PR also.

capture/README.md
nasahlpa: If you want to store the traces in projects/ you need to make sure that the dir. exists as the script is not creating a directory.

capture/capture_aes.py
```diff
@@ -0,0 +1,380 @@
...
+    key = key_fixed
+    logger.info(f"Initializing OT AES with key {binascii.b2a_hex(bytes(key))} ...")
+    ot_aes.write_key(key)
+
+    # Generate plaintexts and keys for first batch.
+    if capture_cfg.batch_mode and capture_cfg.capture_mode == "aes_fsvr_key":
+        ot_aes.write_fvsr_batch_generate(capture_cfg.num_segments.to_bytes(4, "little"))
+
+    if capture_cfg.batch_mode and capture_cfg.capture_mode == "aes_random":
+        ot_aes.write_init_text(text)
```
vogelpi, line 231: Maybe do a single `if batch mode` and then nest the capture mode? Also the definition of the initial plaintext (`text` variable) could go here?

vogelpi (commented):
Thanks @nasahlpa for working on this. I've now also reviewed the `capture_aes.py` script. I have a couple of mostly minor comments.

In addition, I think we should also re-enable aes-fvsr-key non-batch! because this is very useful for debugging issues related to batch captures.

Also, we did have an `optimize_cw_capture()` function in the old `capture.py` responsible for maintaining reasonable batch capture rates for longer running captures. Without this, the capture rate drops very significantly and to below non-batch mode when the CW project format. Can you please re-include this functionality?

capture/configs/aes_sca_cw310.yaml
```diff
@@ -1,9 +1,12 @@
...
+  fpga_bitstream: "../objs/lowrisc_systems_chip_earlgrey_cw310_0.1.bit"
   force_program_bitstream: False
-  fw_bin: "objs/aes_serial_fpga_cw310.bin"
+  fw_bin: "../objs/aes_serial_fpga_cw310.bin"
+  pll_frequency: 100000000
   baudrate: 115200
   output_len_bytes: 16
+husky:
   pll_frequency: 100000000
   target_clk_mult: 0.1
```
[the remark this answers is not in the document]
johannheyszl: If the scope is configured in cycles, those values need to be handed to the scope - hence hey are there I suppose. This is why I didn't like the cfg in cycles for the scope so much ;)

capture/capture_aes.py
```diff
@@ -0,0 +1,380 @@
...
+    Supports four different capture types:
+    * aes_random: Fixed key, random plaintext. TODO HOW ptx generated.
+    * aes_random_batch: Fixed key, random plaintext in batch mode.
+    * aes_fvsr: Fixed vs. random key.
+    * aes_fvsr_batch: Fixed vs. random key batch.
+
+    Args:
+        ot_aes: The OpenTitan AES communication interface.
+        expected_last_ciphertext: The pre-computed ciphertext.
+        ciphertext_len: The length of the ciphertext in bytes.
```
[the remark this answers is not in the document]
nasahlpa: I am not sure what you mean with outdated. The `check_ciphertext` function uses the `ot_aes `communication interface to receive the last ciphertext and compare it against `expected_last_ciphertext`.

capture/configs/aes_sca_cw310.yaml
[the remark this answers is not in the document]
nasahlpa: As @johannheyszl mentioned, the reason for having these parameters is that they are required for the scope setup.

capture/capture_aes.py
```diff
@@ -0,0 +1,380 @@
...
+            # Generate this key from the PRNG.
+            key = bytearray(key_length)
+            for i in range(0, key_length):
+                key[i] = random.randint(0, 255)
+        # Always generate this plaintext from PRNG (including very first one).
+        plaintext = bytearray(16)
+        for i in range(0, 16):
+            plaintext[i] = random.randint(0, 255)
+        # Compute ciphertext for this key and plaintext.
+        cipher = AES.new(bytes(key), AES.MODE_ECB)
```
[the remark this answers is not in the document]
nasahlpa: The old capture script uses the scared.aes library for this purpose, as shown [here](https://github.com/lowRISC/ot-sca/blob/ce04758ecacc08ea9d55760e24dab8d5de817af9/cw/capture.py#L560). However, when I use this library instead of `AES.new(bytes(key), AES.MODE_ECB)`, the capture rate significantly drops. So I keep it for now and and add a TODO if we encounter problems with the capture rate.

nasahlpa:
Thanks for the comments @vogelpi. I've addressed your comments, implemented non-batch `fvsr_key `mode, and added the `optimize_capture` functionality. If OK for you, I'll merge into master.

capture/capture_aes.py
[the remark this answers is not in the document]
vogelpi: Sorry, I meant the list of arguments isn't up-to-date:
- the `scope`, `capture_cfg`, `project`, `cwtarget` arguments of `capture()` arent' mentioned/explained here
- `expected_last_ciphertext` and `ciphertext_len` are not arguments of the function `capture()`

capture/configs/aes_sca_cw310.yaml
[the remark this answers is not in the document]
vogelpi: I thought we agreed that this should not be part of the scope code, no?

What I understood was that 1) we specify number of target clock cycles and the target clock frequency, 2) we define the sampling frequency of the scope (either by reading the configured value or by computing it for Husky based on other settings), 3) and then compute what the resulting offsets are which is still not related to the scope.

The scope code should then just get the sampling rate + the window length and offset. Should we set up a quick meeting to avoid any future misunderstandings and confusion?

capture/capture_aes.py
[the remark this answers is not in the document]
vogelpi: Ah, this is perfect then. Thanks for investigating!

vogelpi:
> Thanks for the comments @vogelpi. I've addressed your comments, implemented non-batch `fvsr_key` mode, and added the `optimize_capture` functionality. If OK for you, I'll merge into master.

Thanks, @nasahlpa . I see you also rebased this on top of master already. Can you please make sure the file [ci/ci_trace_check/golden_traces/aes_sca_random_golden_cw310.db](https://github.com/lowRISC/ot-sca/compare/317425ef3c85fc17951249dc9f2235a01d62e557..9e715fb3b40e8973dba4df192648b1cfe638a3f6#diff-e09a99d95068c7e7230e6fe7ae6bb3fdc194f7fd81687fa0d4008807c6ad9584) is stored in Git LFS?

nasahlpa:
> > Thanks for the comments @vogelpi. I've addressed your comments, implemented non-batch `fvsr_key` mode, and added the `optimize_capture` functionality. If OK for you, I'll merge into master.
> 
> Thanks, @nasahlpa . I see you also rebased this on top of master already. Can you please make sure the file [ci/ci_trace_check/golden_traces/aes_sca_random_golden_cw310.db](https://github.com/lowRISC/ot-sca/compare/317425ef3c85fc17951249dc9f2235a01d62e557..9e715fb3b40e8973dba4df192648b1cfe638a3f6#diff-e09a99d95068c7e7230e6fe7ae6bb3fdc194f7fd81687fa0d4008807c6ad9584) is stored in Git LFS?

Done, the file is stored in git LFS.

capture/scopes/chipwhisperer/husky.py
vogelpi: We quickly discussed offline that in the future the user should define `target_clk_freq` and `sampling_rate` and that helper function should then do the magic and throw an error / adjust + warn if the resulting `adc_mul` isn't integer. We've also said that this can be done in a follow-up PR of this one.

capture/configs/aes_sca_cw310.yaml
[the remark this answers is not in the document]
vogelpi: @johannheyszl , @nasahlpa and I just had a quick discussion and agreed that the scope setup code should indeed just get the sampling rate, window length and offset (both in terms of samples). But the user should be able to specify these values in terms of target clock cycles. The function doing the conversion can be a library function called in the test code. This also allows easily manually overriding the values passed to the scope.

A helper function should be created to extract the current sampling rate from the scope and feed that into the conversion function.

We've also said that this functionality can be added in a separate PR to no further delay this one.

vogelpi (approved):
We just had a quick discussion regarding the two remaining open points (sampling rate control for Husky and relevant parameters, scope setup). Both can be done as a follow-up. So this LGTM. Please make sure CI passes before merging this.

Thanks for the hard work @nasahlpa and your reviews everybody!

capture/configs/aes_sca_cw310.yaml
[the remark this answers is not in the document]
vogelpi: BTW: with the latest bitstream you have to adjust this from 0.1 to 0.24. Maybe that's why CI is currently failing.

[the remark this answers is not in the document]
nasahlpa: Ah thanks, yes this was the issue.

PR #211, opened 2023-11-06
[ci] Add OTBN horizontal measurements to CI

vrozic:
Adds measurements of ECDSA-256 and ECDSA-384 to the CI. Tests both simple and stream mode.


nasahlpa (approved):
LGTM, thanks.

PR #218, opened 2023-11-15
[capture] Add uJSON support

nasahlpa:
This PR enables communication with the OT AES over uJSON.

How to use:
- In `capture/configs/aes_sca_cw310.yaml` `target` change:
  - `protocol: "ujson"` 
  - `port: "/dev/ttyACM4"` (adapt the port accordingly, needs to be UART of CW310)
  - `fw_bin: "../objs/aes_ujson_fpga_cw310.bin"`

Device code is located in [PR#20386](https://github.com/lowRISC/opentitan/pull/20386).


capture/lib/ot_communication.py
```diff
@@ -83,4 +189,52 @@ def read_ciphertext(self, len_bytes):
         Returns:
             The received ciphertext.
         """
-        return self.target.simpleserial_read("r", len_bytes, ack=False)
+        if self.simple_serial:
+            return self.target.simpleserial_read("r", len_bytes, ack=False)
+        else:
+            while True:
+                read_line = str(self.ser.readline())
```
nasahlpa, line 196: [@milesdai] (https://github.com/milesdai) how could this done more efficiently? Basically, I want to read the ciphertext back from OT.
- There might be some other data which is not yet read, therefore I am doing this in a loop
- I am receiving `b'\x11RESP_OK:{"ciphertext":[22,104,44,54,253,255,255,255,52,0,0,0,0,0,0,0,0,64,0,0,128,24,0,0,0,0,0,0,0,0,0,0,0,0,66,64,106,108,0,32,160,254,1,16,16,0,0,0,152,32,0,16,198,41,0,32,132,254,1,16,202,192,0,32],"ciphertext_length":4} CRC:3671789451\n'`. Is there a better way of parsing this instead of using split?

capture/lib/ot_communication.py
```diff
@@ -3,77 +3,183 @@
...
+                self.ser = serial.Serial(port)
+                self.ser.baudrate = baudrate
+                # Init the AES core.
+                self.ser.write(json.dumps("AesSca").encode("ascii"))
+                self.ser.write(json.dumps("Init").encode("ascii"))
+
+    def _ujson_aes_sca_cmd(self):
+        # TODO: without the delay, the device uJSON command handler program
+        # does not recognize the commands.
+        time.sleep(0.2)
```
nasahlpa, line 36: [@milesdai](https://github.com/milesdai) is there a way to avoid this?

capture/lib/ot_communication.py
milesdai: It looks like this code has been revised now, but I think using the json parsing is correct. uJSON will serialize datatypes as valid json after the `RESP_OK:`.

capture/lib/ot_communication.py
milesdai: Just a quick update: I'm working on reproducing this bug. Should I be running something like `./capture_aes.py -c configs/aes_sca_cw310.yaml -p test_project` to test this?

nasahlpa: Yes, exactely. Please change the `configs/aes_sca_cw310.yaml` as mentioned above, build the firmware from the PR linked above, and run `./capture_aes.py -c configs/aes_sca_cw310.yaml -p test_project`. However, if you don't have a ChipWhisperer Husky, the capture script fails.
I'll create a small POC script for you.

nasahlpa: Here is a small POC script:
```
#!/usr/bin/env python3
# Copyright lowRISC contributors.
# Licensed under the Apache License, Version 2.0, see LICENSE for details.
# SPDX-License-Identifier: Apache-2.0

from lib.ot_communication import OTAES, OTPRNG, OTUART

# Establish UART for uJSON command interface. Returns None for simpleserial.
ot_uart = OTUART(protocol="ujson", port="/dev/ttyACM4")

# Create communication interface to OT AES.
ot_aes = OTAES(target=None, protocol="ujson", port=ot_uart.uart)

# Set the key.
key = [0x81, 0x1E, 0x37, 0x31, 0xB0, 0x12, 0x0A, 0x78, 0x42, 0x78, 0x1E, 0x22, 0xB2, 0x5C, 0xDD, 0xF9]
ot_aes.key_set(key)

# Start encryption.
text = [0xAA, 0xAA, 0xAA, 0xAA, 0xAA, 0xAA, 0xAA, 0xAA, 0xAA, 0xAA, 0xAA, 0xAA, 0xAA, 0xAA, 0xAA, 0xAA]
ot_aes.single_encrypt(text)

# Read ciphertext.
print(ot_aes.read_ciphertext(16))
```

Please place this script into the `capture/` folder, transfer the firmware to the device, adapt the port (e.g., `"/dev/ttyACM4"`) in the script, and start it. You should receive the ciphertext.

When I now remove all `time.sleep` commands in `ot_communication.py`, I do not receive a ciphertext. I did some `printf` debugging on the device, it seems that when sending two commands subsequently, the latter is not received by the command handler.

milesdai: Thanks for the test script! I'm able to reproduce the problem you're describing. I'm still looking into this, but so far, this sounds a lot like a flow control issue. The firmware enables software-based flow control through OTTF, so I've tried enabling SW flow control on the Python side as well with the [`xonxoff=True` flag](https://pyserial.readthedocs.io/en/latest/pyserial_api.html#serial.Serial.__init__), but no luck so far.

I'm going to take a closer look at this, but in the meanwhile, I've found that I can reduce the sleep time to about 10 ms instead of 200. If we don't want to block the PR on this, maybe that can help your throughput a bit.

nasahlpa: Thanks @milesdai! I'll decrease the timeout and merge the PR for now.
Please let me know if I can help you with debugging, for us a fast communication is really important as we want to trigger AES executions as fast as possible to achieve high capture rates.

nasahlpa:
As discussed with @milesdai , capture rate is still quite low as there seems to be an issue with the communication protocol. Miles will be working on a fix. 
To not block progress on this repo, could others (@vogelpi @johannheyszl @vrozic ...) please review this PR?

vogelpi (commented):
Thanks @nasahlpa , the PR looks mostly good but especially in `data_generator.py` there seem to be changes not related to this PR.

vrozic:
> Capture rates for AES random batch drops from 934.27 traces/s to 350.40 traces/s.

This is probably not a problem. If we increase `num_segments` to e.g. 160, the capture rate goes back up to ~2500. I think that for these measurement on Husky we can use up to 190 segments.

capture/lib/ot_communication.py
```diff
@@ -3,86 +3,223 @@
...
+            self._ujson_aes_sca_cmd()
+            # KeySet command.
+            self.port.write(json.dumps("KeySet").encode("ascii"))
+            # Key payload.
+            time.sleep(0.01)
+            key_data = {"key": key, "key_length": key_length}
+            self.port.write(json.dumps(key_data).encode("ascii"))
 
-    def fvsr_key_set(self, key):
+    def fvsr_key_set(self, key, key_length: Optional[int] = 16):
```
vrozic, line 63: It looks like this function is not used anywhere.

nasahlpa: Yes, you are right. However, as the AES SCA uJSON handler at the OT device implements the _FvsrKeySet_ command, the handler is provided here. If we don't need the _FvsrKeySet_ command at the OT device, we should remove it there and also here.

vrozic (approved):
I've tested this PR and all captures work well with my setup.

util/data_generator.py
```diff
@@ -2,60 +2,59 @@
 # Licensed under the Apache License, Version 2.0, see LICENSE for details.
 # SPDX-License-Identifier: Apache-2.0
 
-import numpy as np
```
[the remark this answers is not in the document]
vogelpi: Ah, thanks for explaining Pascal. This makes sense. I thought it was only cosmetic at first :-)

vogelpi (approved):
LGTM!

vogelpi:
> > Capture rates for AES random batch drops from 934.27 traces/s to 350.40 traces/s.
> 
> This is probably not a problem. If we increase `num_segments` to e.g. 160, the capture rate goes back up to ~2500, which is still pretty high. I think that for these measurement on Husky we can use up to 190 segments.

Related to that, aren't the number of segments automatically determined by Husky? I thought this was required to always completely fill the sample memory in Husky before reading it out in batch mode. Do we manually set the number of segments somewhere now?

vogelpi:
> > Related to that, aren't the number of segments automatically determined by Husky?
> 
> That used to be the case. Now it's just specified in the configuration file. For this reason we also sometimes get more measurements than we ask for (e.g. if num_segments = 40, num_traces = 100, the script will take 120 measurements)

Ah, thanks for getting back! This makes sense. We should probably add a check in case num_segements * num_samples is bigger than the sample memory. Pascal said to take care of this.

PR #219, opened 2023-11-20
Aes capture dtr

vrozic:
This PR makes the following changes related to AES capture

1. Acknowledges `batch_encrypt` command to resolve timing issues that I was seeing
2. Adds `data_generator.py` module to util. This module is used to generate plaintexts and keys compliant with Derived Test Requirements. For the time being this module is used only for AES captures but it can be reused for KMAC and SHA3 captures.
3. Updates binary and integrates data_generator into `capture_aes.py`


util/data_generator.py
```diff
@@ -0,0 +1,71 @@
...
+
+    def set_start(self):
+        self.text_fixed = bytearray([0xAA, 0xAA, 0xAA, 0xAA, 0xAA, 0xAA, 0xAA, 0xAA,
+                                     0xAA, 0xAA, 0xAA, 0xAA, 0xAA, 0xAA, 0xAA, 0xAA])
+        self.text_random = bytearray([0xCC, 0xCC, 0xCC, 0xCC, 0xCC, 0xCC, 0xCC, 0xCC,
+                                      0xCC, 0xCC, 0xCC, 0xCC, 0xCC, 0xCC, 0xCC, 0xCC])
+        self.key_fixed = bytearray([0x81, 0x1E, 0x37, 0x31, 0xB0, 0x12, 0x0A, 0x78,
+                                    0x42, 0x78, 0x1E, 0x22, 0xB2, 0x5C, 0xDD, 0xF9])
+        self.key_random = bytearray([0x53, 0x53, 0x53, 0x53, 0x53, 0x53, 0x53, 0x53,
+                                     0x53, 0x53, 0x53, 0x53, 0x53, 0x53, 0x53, 0x53])
```
nasahlpa, line 36: Is there a particular reason why the initial and fixed values are here hardcoded? Previously, we used the fixed and init. crypto material from the config file.

Then, we should add a comment in the config files that this crypto material is only used for the random test but not fvsr.

nasahlpa (approved):
Thanks @vrozic, LGTM!

util/data_generator.py
vrozic: Yes, DTR prescribes these starting values.
For now, I will add the comment in the config file, but in perspective I will also change random capture to use DTR values.

nasahlpa: Makes sense, thanks for the explanation!

PR #220, opened 2023-11-20
[scope] Autoconvert cycles into samples

nasahlpa:
Previously, the conversion from cycles into samples happened in the Husky scope driver. This PR moves this conversion into a library, allowing the user to either specify offset and windowlength in samples or cycles.
In addition, the sampling rate also can be changed in the config file.

In the future, we want to read the sampling rate from the scope.

### DIFF
--- a/capture/capture_aes.py
+++ b/capture/capture_aes.py
@@ -19,7 +19,8 @@ import yaml
 from Crypto.Cipher import AES
 from lib.ot_communication import OTAES
 from project_library.project import ProjectConfig, SCAProject
-from scopes.scope import Scope, ScopeConfig
+from scopes.cycle_converter import convert_num_cycles, convert_offset_cycles
+from scopes.scope import Scope, ScopeConfig, determine_sampling_rate
 from tqdm import tqdm
 
 sys.path.append("../")
@@ -66,6 +67,11 @@ def setup(cfg: dict, project: Path):
     Returns:
         The target, scope, and project.
     """
+    # Calculate pll_frequency of the target.
+    # target_freq = pll_frequency * target_clk_mult
+    # target_clk_mult is a hardcoded constant in the FPGA bitstream.
+    cfg["target"]["pll_frequency"] = cfg["target"]["target_freq"] / cfg["target"]["target_clk_mult"]
+
     # Init target.
     logger.info(f"Initializing target {cfg['target']['target_type']} ...")
     target = CWFPGA(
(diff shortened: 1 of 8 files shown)


capture/capture_aes.py
```diff
@@ -80,15 +81,30 @@ def setup(cfg: dict, project: Path):
     # Init scope.
     scope_type = cfg["capture"]["scope_select"]
     logger.info(f"Initializing scope {scope_type} ...")
+    # Convert number of cycles into number of samples, if necessary.
```
vogelpi, line 84: For Husky, we should check that the sampling_rate is indeed an integer multiple of pll_frequency. And print a warning if it's not. Otherwise, the scope is configured with wrong num_samples and offset_samples values.

capture/configs/aes_sca_cw310.yaml
```diff
@@ -4,11 +4,11 @@ target:
   force_program_bitstream: True
   fw_bin: "../objs/aes_serial_fpga_cw310.bin"
   pll_frequency: 100000000
+  target_freq: 24000000
```
vogelpi, line 7: In my view, we should have the target_clk_mult value instead of the pll_frequency value here. And we should say in a comment directly above target_clk_mult that target_clk_mult is not to be changed by the user but that this is a value hard-coded in the bitstream. The user should then configure the target_freq value, and based on this value and the  target_clk_mult  we internally determine the pll_frequency and afterwards the sampling_rate (where sampling_rate = pll_frequency * int).

I think there is little value in being able to specify both pll_frequency and target_freq because there is a fixed relationship between the two.

Does this sound reasonable or is there a better way to handle this maybe?

vogelpi (commented):
Thanks @nasahlpa for doing this!

The implementation looks good to me modulo two things:
- I think we should print a warning when the sampling rate isn't an integer multiple of the pll_frequency, and accordingly tune the sample values.
- The user should specify target_freq only. Other values like pll_frequency should be derived internally based on the board-specific target_clk_mult value (not to be changed by the user).

capture/configs/aes_sca_cw310.yaml
nasahlpa: As all three values are related, e.g., target_freq

vogelpi: The only value fixed by the bitstream is `target_clk_mult`. `pll_frequency` is what we configure via the CW API and `target_freq` is the resulting target frequency determined by your equation.

What I am trying to avoid is that people for whatever reason configure e.g. 50 MHZ `pll_frequency` and 10 MHz `target_freq`, and then spend time debugging why it's not working as expected. It should be crystal clear for users which value to tune and which values to leave alone.

As for the sampling rate: It cannot arbitrarily be chosen on Husky. There is always the fixed relationship `sampling_rate` = `pll_frequency * adc.mul`  where `adc.mul` must be an integer. So everything is kind of linked together. Luckily, we have more freedom with other scopes. Again, what I am trying to avoid is that people configure values that don't work and then wonder why. We will spend time supporting them, debugging their issues, and explain to them how things work.

nasahlpa: Make sense, `target_clk_mult` and `target_freq `are now configurable.

capture/capture_aes.py
nasahlpa: I added a check to make sure that `adc.mul` is an integer

capture/configs/aes_sca_cw310.yaml
nasahlpa: @vogelpi now it should be implemented as discussed. `sampling_rate `can be optional, then the scope driver automatically calculates the maximum available sampling rate and takes care that adc.mul is a int.

vogelpi: This is perfect, thanks @nasahlpa !

vogelpi (approved):
LGTM!

PR #222, opened 2023-11-21
[capture] Set num_segments in non-batch mode

nasahlpa:
As discovered by @m-temp, when running a non-batch capture but having num_segments set to > 1, capture fails. This PR fixes the issue by forcefully setting num_segments=1.


vogelpi (approved):
LGTM!

m-temp:
Thx @nasahlpa for helping me on this.

I'm not a fan of forcefully setting/changing/ignoring config values, as this can cause quite some time for debugging.
I've added an assertion in the kmac script, however we should align these things. If an assertion is too strict, what about an warning, so that the user at least knows that the parameter was overwritten?

nasahlpa:
> Thx @nasahlpa for helping me on this.
> 
> I'm not a fan of forcefully setting/changing/ignoring config values, as this can cause quite some time for debugging. I've added an assertion in the kmac script, however we should align these things. If an assertion is too strict, what about an warning, so that the user at least knows that the parameter was overwritten?

Good point, I agree.

I would propose to print a warning and change `num_segments=1`.

PR #224, opened 2023-11-22
Add initial FI infrastructure

nasahlpa:
This PR pushes the first code for the FI penetration infrastructure.

The most important functionality is the parameter sweep function:
```python
remaining_iterations = cfg["fisetup"]["num_iterations"]
with tqdm(total=remaining_iterations, desc="Injecting", ncols=80, unit=" different faults") as pbar:
    while remaining_iterations > 0:
        # Get fault parameters (e.g., trigger delay, glitch voltage).
        fault_parameters = fi_gear.generate_fi_parameters()

        # Arm the FI gear.
        fi_gear.arm_trigger(fault_parameters)

        # Start test on OpenTitan.
        ot_communication.start_test(cfg)

        # Read response.
        response = ot_communication.read_response()

        # Compare response.
        # Check if result is expected result (FI failed), unexpected result
        # (FI successful), or no response (FI failed.)
        fi_result = FISuccess.SUCCESS
        if response == cfg["test"]["expected_result"]:
            # Expected result received. No FI effect.
            fi_result = FISuccess.EXPRESPONSE
        elif response == "":
            # No UART response received.
            fi_result = FISuccess.NORESPONSE

        # Store result into FIProject.
        project.append_firesult(
            response = response,
            fi_result = fi_result,
            trigger_delay = fault_parameters["trig_delay"],
            glitch_voltage = fault_parameters["glitch_voltage"],
            glitch_len = fault_parameters["glitch_len"]
        )

        remaining_iterations -= 1
        pbar.update(1)
```

This function first generates random fault injection parameters within the given limits. For voltage glitching, the following parameters are randomly generated: `glitch_voltage`, `glitch_len`, and `trigger_delay`. These parameters are then handed to the FI gear (voltage glitcher, ...). Then, communciation with OT over uJSON (no simpleserial support) is established and the test is started. The test on OT will raise the trigger and the FI gear injects the fault. Afterwards, the result is fetched from the device and analyzed. Possible options for a response could be: no response, unexpected response, expected response. FI is successful, if an unexpected response is received. The result is stored into the database and a plot over the fault parameter space is generated:
image
This plot helps to narrow down the fault parameters to increase the fault success rate.
Yellow: expected response, red: no response (device probably crashed), and green: unexpected response (potential successful response).

Currently, only a dummy VCC glitcher is supported. Support for other gear (LFI, EMFI, VCC & clock glitching) needs to be added.

Device PR: [[fi] Add Ibex FI command handlers #20496](https://github.com/lowRISC/opentitan/pull/20496)


nasahlpa:
Q for @johannheyszl, what FI parameters do we need?

Voltage glitching:
- `glitch_len`
- `glitch_voltage`
- `trigger_delay`

Clock glitching:
- `glitch_width` (% of clock period)
- `trigger_delay`

EMFI:
- `x_pos`
- `y_pos`
- `voltage`
- `pulse_width`
- `trigger_delay`

LFI:
- `x_pos`
- `y_pos`
- `intensity`
- `pulse_width`
- `trigger_delay`

johannheyszl:
@nasahlpa thanks! Nit:

Voltage glitching:
```
trigger_delay
**glitch_len -> glitch_width**
glitch_voltage
```

LFI:
```
x_pos
y_pos
**z_pos**
trigger_delay
**intensity -> attenuation**
pulse_width
```
Here we will see more sometimes, e.g. if dual lasers are used, and bc besides x-y-table, the beam can also often be x-y controlled separately. But I think we can go with this for now.

nasahlpa:
> glitch_width



> Voltage glitching:
> 
> ```
> trigger_delay
> **glitch_len -> glitch_width**
> glitch_voltage
> ```

Thanks, changed to glitch width.

I will add other driver dummy classes for LFI, EMFI, & clock glitching in a separate PR.

fault_injection/fi_ibex.py
```diff
@@ -0,0 +1,200 @@
...
+
+    Printing the plot helps to narrow down the fault injection parameters.
+
+    Args:
+        project: The project containing the traces.
+        config: The capture configuration.
+        file: The file path.
+    """
+    if config["fiproject"]["show_plot"]:
+        # TODO: implement FI plot.
```
vogelpi, line 147: This function is now implemented no?

fault_injection/fi_ibex.py
```diff
@@ -0,0 +1,200 @@
...
+            # (FI successful), or no response (FI failed.)
+            fi_result = FISuccess.SUCCESS
+            if response == cfg["test"]["expected_result"]:
+                # Expected result received. No FI effect.
+                fi_result = FISuccess.EXPRESPONSE
+            elif response == "":
+                # No UART response received.
+                fi_result = FISuccess.NORESPONSE
+                # Resetting OT as it most likely crashed.
+                cwtarget.reset_target()
```
vogelpi, line 111: Would it make sense to first only try resetting OT instead of directly re-flashing the bitstream?

fault_injection/fi_gear/husky/husky_vcc.py
```diff
@@ -0,0 +1,101 @@
...
+        Returns:
+            A dict containing the FI parameters.
+        """
+        parameters = {}
+        parameters["glitch_width"] = random_float_range(self.glitch_width_min,
+                                                        self.glitch_width_max,
+                                                        self.glitch_width_step)
+        parameters["trigger_delay"] = random_float_range(self.trigger_delay_min,
+                                                         self.trigger_delay_max,
+                                                         self.trig_step)
```
vogelpi, line 93: I am wondering whether there is a smart way to avoid re-running the experiment using the same parameters. May keep them all around in a big array and generate new values until the combination doesn't completely match anymore?

vogelpi: Also, I think we discussed that some parameters like the x,y position maybe doesn't need to be random. Obviously, we don't have this for Husky but maybe we should already add support for non-randomized parameters?

target/communication/fi_ibex_commands.py
```diff
@@ -0,0 +1,164 @@
...
+    def ibex_char_unrolled_reg_op_loop(self) -> None:
+        """ Starts the ibex.char.unrolled_reg_op_loop test.
+        """
+        # IbexFi command.
+        self._ujson_ibex_fi_cmd()
+        # CharUnrolledRegOpLoop command.
+        time.sleep(0.01)
+        self.uart.write(json.dumps("CharUnrolledRegOpLoop").encode("ascii"))
+
+    def ibex_char_unrolled_mem_op_loop(self) -> None:
```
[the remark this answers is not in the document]
nasahlpa: Yeah, you are right. I'll remove the other command handlers and add them when also having the tests in the firmware.

target/cw_fpga.py
```diff
@@ -72,11 +77,15 @@ def __init__(self, bitstream, force_programming, firmware, pll_frequency,
...
-                                         pll_frequency)
+        self.fpga = self.initialize_fpga(self.fpga_type, bitstream,
+                                         force_programming, pll_frequency)
 
         self.target = self.initialize_target(programmer, firmware, baudrate,
                                              output_len, pll_frequency)
+
+        self.target = self.initialize_target(programmer, firmware,
+                                             baudrate, output_len,
+                                             pll_frequency)
```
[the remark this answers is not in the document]
nasahlpa: Argh, this is the result of a merge conflict..

fault_injection/project_library/project.py
```diff
@@ -0,0 +1,120 @@
...
+# Licensed under the Apache License, Version 2.0, see LICENSE for details.
+# SPDX-License-Identifier: Apache-2.0
+
+import sys
+from dataclasses import dataclass
+from enum import IntEnum
+from pathlib import Path
+from typing import Optional
+
+sys.path.append("../")
```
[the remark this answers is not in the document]
nasahlpa: Basically, this is what I want:
`import ../util.helpers as helpers`
However, as this seems not to be possible in Python, I first change the current path:
`sys.path.append("../")`
and then can find the util.helpers lib:
`import util.helpers as helper`
I am not really sure whether there is a better way of doing this.

fault_injection/fi_gear/husky/husky_vcc.py
nasahlpa: Yeah, this could be an option. However, in my experience, a completely random strategy works also quite well.

Regarding the deterministic fault parameter sweep, I'll create an issue & implement it.

PR #226, opened 2023-11-23
Add CW version check

nasahlpa:
As this repository requires a specific ChipWhisperer API version as well as ChipWhisperer Husky firmware version, this PR adds a version check to both.


vrozic (approved):
Thanks @nasahlpa for working on this.
This looks good to me.

PR #230, opened 2023-11-28
Pin Python requirements to version

nasahlpa:
As discovered by @johannheyszl, some sqlalchemy versions cause issues when fetching data from the DB. This PR pins all Python requirements to a specifc version. The versions are identical to what currently is used in the podman container.


johannheyszl:
Thanks a lot! This helped :)

PR #231, opened 2023-11-28
Create project dir automatically

nasahlpa:
Previously, an error was thrown when the project dir (-p argument) was not provided. Now, instead a warning is printed and the dir is created.


johannheyszl (approved):
nice thanks!

PR #232, opened 2023-11-28
Fetch path of plot from cmd line args

nasahlpa:
In this PR, the path for the plot is taken from the -p command line argument instead from the config file. This streamlines the usage of the script.


johannheyszl (approved):
Thanks!

PR #244, opened 2023-12-05
[capture] Enable uJSON for KMAC

### LINKED ISSUE #237: [ci] Add KMAC to CI
With the new KMAC capture script, we should enable CI for KMAC.

### LINKED ISSUE #238: [capture] Enable uJSON for KMAC
Similar to AES (#218), we need to enable uJSON communication for KMAC. This requires adaptions in the ot_communication file as well as on Opentitan ([#20386](https://github.com/lowRISC/opentitan/pull/20386)).

nasahlpa:
Similar to #218 , this PR enables uJSON support for capturing KMAC traces. The device command handler code can be found in [#20563](https://github.com/lowRISC/opentitan/pull/20563).

Consists of three commits:
- uJSON implementation. Closes #238
- Adds KMAC to CI. Closes #237 
- Switch byte array to int array in data generator.

### DIFF
--- a/util/data_generator.py
+++ b/util/data_generator.py
@@ -2,9 +2,16 @@
 # Licensed under the Apache License, Version 2.0, see LICENSE for details.
 # SPDX-License-Identifier: Apache-2.0
 
-import numpy as np
 from Crypto.Cipher import AES
 from Crypto.Hash import KMAC128
+
+"""Data generator.
+
+Generates crypto material for the SCA tests.
+
+Input and output data format of the crypto material (ciphertext, plaintext,
+and key) is plain integer arrays.
+"""
 
 
 class data_generator():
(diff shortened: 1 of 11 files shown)


util/data_generator.py
```diff
@@ -62,21 +67,21 @@ def get_random(self):
...
 
     def get_kmac_fixed(self):
-        pt = np.asarray(self.text_fixed)
-        key = np.asarray(self.key_fixed)
+        pt = self.text_fixed
+        key = self.key_fixed
         mac_fixed = KMAC128.new(key=bytes(self.key_fixed), mac_len=32)
         mac_fixed.update(bytes(self.text_fixed))
-        ct = np.asarray(bytearray(mac_fixed.digest()))
+        ct = mac_fixed.digest()
```
m-temp, line 74: I'm not a python expert and also got confused with the different byte representations, but I'm not sure if the byte/int/array problematic is fully covered here.

Imho `.update()` method requires byte strings or byte arrays and also returns a byte array. Are they dynamically casted?

Further, we later only use the ct to compare it in with the received one. AFAIK this also is done in bytearrays. So we should be fine here. Anyways, I guess it would be good to document where we use which kind of data representation. I'd suggest to put this information at the top of each script.

m-temp (approved):

m-temp:
Regarding the CI implementation: I've never done that, so I cannot comment on that

PR #248, opened 2023-12-08
[capture] Enable uJSON for SHA3

nasahlpa:
Similar to #218 , this PR enables uJSON support for capturing SHA3
traces. The device command handler code can be found in [#20593](https://github.com/lowRISC/opentitan/pull/20593).

### DIFF
--- a/capture/lib/ot_communication.py
+++ b/capture/lib/ot_communication.py
@@ -39,7 +39,7 @@ class OTAES:
 
     def _ujson_aes_sca_cmd(self):
         # TODO: without the delay, the device uJSON command handler program
-        # does not recognize the commands.
+        # does not recognize the commands. Tracked in issue #256.
         time.sleep(0.01)
         self.port.write(json.dumps("AesSca").encode("ascii"))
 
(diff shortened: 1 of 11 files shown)


capture/lib/ot_communication.py
```diff
@@ -404,62 +404,153 @@ def read_ciphertext(self, len_bytes):
...
+                json_string = read_line.split("RESP_OK:")[1].split(" CRC:")[0]
+                try:
+                    status = json.loads(json_string)["status"]
+                    if status[0] != 0:
+                        raise Exception("Batch mode acknowledge error: Device and host not in sync")
+                    return status
+                except Exception:
+                    raise Exception("Batch mode acknowledge error: Device and host not in sync")
+            else:
+                raise Exception("Batch mode acknowledge error: Device and host not in sync")
```
m-temp, line 441: There are two commands where `_ujson_sha_sca_ack()`  is called: `set_mask_off/on()` and `absorb_batch()`.
Thus, the error message might be missleading if this error happens in non-batch mode for the mask on/off function.
I think it is sufficient to just say "Acknowledge error:..." and use the exception information to track the function where it occurred.

m-temp (approved):
The ujson part looks good to me. With the CI I'm not familiar.
There a couple of `time.sleep(0.01)`. I remember that there is (was ?) an issue that also deals with sleep commands, but I don't find it atm. Is this related and if so, should we track this to get rid of the sleeps in the future?

capture/lib/ot_communication.py
nasahlpa: Good point, I've changed the message.

PR #259, opened 2023-12-13
[fi] Add Ibex FI characterization handlers

nasahlpa:
Adds command handlers for the following tests used to characterize Ibex:
- ibex.char.mem_op_loop
- ibex.char.reg_op_loop
- ibex.char.unrolled_mem_op_loop

Device code is located in [#20649](https://github.com/lowRISC/opentitan/pull/20649).


vogelpi (approved):
LGTM!

PR #260, opened 2023-12-14
[fi] Reset OpenTitan on faulty behavior

### LINKED ISSUE #249: [fi] Reset OT on CW310
Currently, when OT is not responding, we re-program the bitstream on CW310. As this takes some time, we could:

- Check whether the bitstream on CW310 is still ok (CRC check)
- And only reset OT if it is OK

nasahlpa:
Previously, we re-programmed the FPGA when a FI crashed OpenTitan running on CW310. As this took quite some time, we now read the INITB pin - this pin returns false or none when the FPGA is not programmed or the bitstream CRC does not match.
In this case, we are re-programming the bistream.

When the bitstream is OK, we simply reset OpenTitan by using the power on reset pin (POR_N is mapped to USB_A14 on CW310).

This significantly increases the FI rate.

Closes #249.

### DIFF
--- a/target/cw_fpga.py
+++ b/target/cw_fpga.py
@@ -189,17 +189,35 @@ class CWFPGA(object):
         time.sleep(0.5)
 
     def reset_target(self):
-        """Resets the target by re-programming the FPGA. """
-        programmer = SpiProgrammer(self.fpga_type)
+        """Resets the target. """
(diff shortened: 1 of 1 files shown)


vogelpi (approved):
Excellent, thanks @nasahlpa !

PR #262, opened 2023-12-18
[scope] Check num_segments_max for CW Husky

### LINKED ISSUE #247: Husky sample memory
As `num_segments` can be defined by the user in the config file, we should add a check ensuring that `num_segements * num_samples < sample memory`.

nasahlpa:
As the user can manually define num_segments in the config file, we need to make sure that we can keep num_segments*num_samples in the CW Husky sample memory.

Closes #247.


vogelpi (approved):
LGTM, thanks @nasahlpa !

PR #263, opened 2023-12-18
[db] Add notes to database

### LINKED ISSUE #213: [capture] Allow user to enter notes via CLI
Currently, the notes stored in the trace database need to be added in the Python script. In a future version of the capture script, the notes could be passed as an argument when executing the script or during the execution in the CLI.

nasahlpa:
This PR allows the user to pass notes as an argument to the project database.

Closes #213.


vogelpi (approved):
Very nice, thanks @nasahlpa !

PR #264, opened 2023-12-18
[db] Store binary blob into database

### LINKED ISSUE #214: [capture] Store binary into database
Currently, only the path to the used binary is stored into the trace database. In a future version of the capture script, the binary itself could be stored in the database.

nasahlpa:
This PR enables the user to save the bitstream and the used binary into the project database. Moreover, a CRC of the used binary and bistream is calculated and stored into the database.

Closes #214.

### DIFF
--- a/fault_injection/fi_ibex.py
+++ b/fault_injection/fi_ibex.py
@@ -178,12 +178,21 @@ def main(argv=None):
     # Save metadata.
     metadata = {}
     metadata["datetime"] = datetime.now().strftime("%m/%d/%Y, %H:%M:%S")
-    metadata["cfg"] = cfg
-    metadata["fpga_bitstream"] = cfg["target"]["fpga_bitstream"]
-    # TODO: Store binary into database instead of binary path.
-    # (Issue lowrisc/ot-sca#214)
-    metadata["fw_bin"] = cfg["target"]["fw_bin"]
+    # Store bitstream information.
+    metadata["fpga_bitstream_path"] = cfg["target"]["fpga_bitstream"]
+    metadata["fpga_bitstream_crc"] = helpers.file_crc(cfg["target"]["fpga_bitstream"])
+    if args.save_bitstream:
+        metadata["fpga_bitstream"] = helpers.get_binary_blob(cfg["target"]["fpga_bitstream"])
+    # Store binary information.
+    metadata["fw_bin_path"] = cfg["target"]["fw_bin"]
+    metadata["fw_bin_crc"] = helpers.file_crc(cfg["target"]["fw_bin"])
+    if args.save_binary:
+        metadata["fw_bin"] = helpers.get_binary_blob(cfg["target"]["fw_bin"])
+    # Store user provided notes.
     metadata["notes"] = args.notes
+    # Store the Git hash.
+    metadata["git_hash"] = helpers.get_git_hash()
+    # Write metadata into project database.
     project.write_metadata(metadata)
 
     # Save and close project.
--- a/util/helpers.py
+++ b/util/helpers.py
@@ -3,7 +3,52 @@
 # SPDX-License-Identifier: Apache-2.0
 
 import argparse
+import zlib
 from pathlib import Path
+
+import git
+
+
+def get_git_hash() -> str:
+    """Get the Git hash of the repository
+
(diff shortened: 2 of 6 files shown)


util/helpers.py
```diff
@@ -3,9 +3,42 @@
...
+def file_crc(file_path: Path) -> str:
+    """Calculates CRC32 of the provided file
+
+    Args:
+        file_path: The path of the file to calculate the CRC32.
+
+    Returns:
+        String containing the CRC32 of the file.
+    """
+    crc_str = 0
```
vogelpi, line 19: The naming of the variable is ambiguous. `zlib.crc32()` [returns an unsigned 32-bit integer](https://docs.python.org/3/library/zlib.html#zlib.crc32).

util/helpers.py
```diff
@@ -60,6 +93,20 @@ def parse_arguments(argv):
...
+                        required=False,
+                        help="Save bitstream into project database")
+
+    parser.add_argument("-f",
+                        "--save_binary",
+                        dest="save_binary",
+                        type=bool,
+                        required=False,
+                        help="Save binary into project database")
+
```
vogelpi, line 109: In my view, I would always save both or none of them. Or do you have a specific scenario in mind where only one is useful?

fault_injection/fi_ibex.py
```diff
@@ -178,11 +178,16 @@ def main(argv=None):
...
     metadata = {}
     metadata["datetime"] = datetime.now().strftime("%m/%d/%Y, %H:%M:%S")
-    metadata["cfg"] = cfg
-    metadata["fpga_bitstream"] = cfg["target"]["fpga_bitstream"]
-    # TODO: Store binary into database instead of binary path.
-    # (Issue lowrisc/ot-sca#214)
-    metadata["fw_bin"] = cfg["target"]["fw_bin"]
+    # Store bitstream information.
+    metadata["fpga_bitstream_path"] = cfg["target"]["fpga_bitstream"]
+    metadata["fpga_bitstream_crc"] = helpers.file_crc(cfg["target"]["fpga_bitstream"])
```
vogelpi, line 183: I like the idea of the CRC!

What I wondering is if it is of any use if we're not storing the binary/bitstream? Or do you envision a use case where one uses e.g. the bistream currently in the repo but experiments with different binaries? And then when going back later one can use the bistream that was in the repo back then?

I think this could work, because in most cases one will have to go back in time in the repo anyway to reproduce results (to do that the Git Hash would be nice! I don't know if we're currently doing that).

But just from a perspective of saving storage, I don't think it's worth it. Bitstream and binary together are maybe 100 MB at most, but the traces themselves are quickly in the GB domain. Or what is your view?

vogelpi (approved):
Thanks @nasahlpa , I like the idea of the CRC. I have some questions regarding combining arguments.

util/helpers.py
nasahlpa: Fixed.

util/helpers.py
nasahlpa: When we do not use an FPGA in the future but still have the capabilities of changing the binary on the discrete chip, this could be useful.

fault_injection/fi_ibex.py
nasahlpa: I implemented this because I know I will make the following in the future:
Because I want to save some storage space for small experiments, I'll omit the `save binary` argument.
When then debugging an issue (with the bitstream or the binary), the CRC field in the database then could be handy as I simply need to compare the output of `crc32 <bitstream/binary>` with the entry.

Git hash is a good idea, I'll quickly implement this.

util/helpers.py
vogelpi: Ah yes, of course. You're right!

fault_injection/fi_ibex.py
vogelpi: Keeping the CRC always makes definitely sense and I see the value for debugging. I am just wondering if we should avoid the possibility of getting into the situation where "debugging" is needed by always storing the binary. But again, when using the ES silicon with the firmware in the chip already, there is no point int storing the binary. So, I agree to keep storing the binaries optional.

Thanks for adding the hash! :+1:

PR #270, opened 2024-01-05
[tvla] Fix plotting issues with new database and metadata format

wettermo:
This commit fixes incompatibility of tvla plotting with the new database and metadata format.

For this, additional metadata is stored in aes, otbn, kmac and sha3 capture. Also, a new wrapper function for plotting is added to the tvla script. Finally, aes, otbn, kmac and sha3 config files are adjusted.

Further, this PR adds code to also include metadata in plots of aes specific tests.

Through this PR two tasks in https://github.com/lowRISC/ot-sca/issues/72 are addressed:
- [x] At the moment, TVLA figures contain meta-data only for general TVLA test. We should enable the same for AES specific TVLA tests.
- [x] At the moment, tvla.py cannot generate some figures if data is loaded from the new database format.

### DIFF
--- a/analysis/configs/tvla_cfg_aes_specific_byte0_rnd0.yaml
+++ b/analysis/configs/tvla_cfg_aes_specific_byte0_rnd0.yaml
@@ -1,4 +1,4 @@
-project_file: projects/simple_capture_aes_sca
+project_file: ../capture/projects/simple_capture_aes_sca
 trace_file: null
 trace_start: null
 trace_end: null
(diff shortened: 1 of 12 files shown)


nasahlpa (approved):
Thanks for adapting all the scripts!

vrozic (approved):
Thanks @wettermo 
This looks good to me.
The CI failure is unrelated to this PR, so I think it can be safely merged.

wettermo:
Thanks @nasahlpa and @vrozic for the reviews!

wettermo:
I'm merging this PR, even though CI fails, because CI failure is unrelated to this PR.

PR #273, opened 2024-01-08
[capture] Move cycle converter into scope.py file

### LINKED ISSUE #243: [capture] Move scope helper functions from cycle-converter to scope?
IMO the two helper functions in [cycle_converter.py ](https://github.com/lowRISC/ot-sca/blob/master/capture/scopes/cycle_converter.py) should also go into scope.py because: they get scope cfg as input, hence, are specific to scope.

We should also rename to them to be consistent with `determine_sampling_rate` in scope.py. Maybe even group them under a function `check_and_convert_cycle_based_cfg`.

Would need some refactor in all already existing capture scripts.

nasahlpa:
As discussed in #243, it makes sense to have the cycle converter helper functions inside the scope.py file.

Closes #243.


johannheyszl (approved):
Thanks @nasahlpa !

Issue #274, opened 2024-01-08
[capture] CW databases broken


vogelpi:
Currently, our tvla script seems to only extract the first 20k traces of a CW database. I have several databases holding 2M and 10M AES traces but only the first 20k traces are effectively read out.

Does anyone know why this is the case and how to fix it? @nasahlpa @vrozic ?

johannheyszl:
oh, back in October/November when I last used the script, it reported on stdout that it would use larger numbers. Does it report 20k on stdout now?

nasahlpa:
I am not aware of such a limit - could you please provide me with the DB file and I try to reproduce & fix it.

vogelpi:
Hm, this is strange. I've now switched to an old version and it still just takes the first 20k traces. I am wondering if maybe the database is actually broken.

vogelpi:
A quick investigation of the generated CW project files reveals that all but the last two 10k segments are "disabled". When storing the segments or at the very end of the capture, we have to enable them. This seems to have gone wrong when reworking the capture setup. I'll adjusted the title accordingly and will file a PR.

PR #275, opened 2024-01-08
[capture] Re-enable all CW trace storage segments before saving to disk

vogelpi:
When using the CW trace database format, we disable all but the last two trace storage segments during captures to maintain high capture rates. Before storing the database to disk at the end of the capture, all trace storage segments need to be re-enabled. Otherwise, only 20k traces are read from the database upon loading.

This resolves lowRISC/ot-sca#274.

### DIFF
--- a/capture/capture_aes.py
+++ b/capture/capture_aes.py
@@ -450,6 +450,8 @@ def main(argv=None):
     # Write metadata into project database.
     project.write_metadata(metadata)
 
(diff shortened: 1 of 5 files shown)


nasahlpa (approved):
Thanks for spotting this, @vogelpi! The enabling of the segments went missing when working on the initial simple scripts.

Issue #283, opened 2024-01-11
[ci] SHA3 captures seem to be broken/flaky in CI


vogelpi:
I've observed that in very many cases, CI captures for SHA3 don't work. We should investigate what is going on an fix it.

nasahlpa:
Yes, I noticed this too for the uJSON version of SHA3. The reason for that is mostly that we wait for an [ACK ](https://github.com/lowRISC/ot-sca/blob/cb90c9c9f6cbd74bafc5818bd6fc8e289fb5336e/capture/lib/ot_communication.py#L427)from OT after sending command from ot-sca.

I am not sure why we introduced this in `simpleserial`? For AES this is also not needed? Maybe @m-temp or @vrozic know more?

m-temp:
I don't know what's happening in the CI, but for SHA3 (and KMAC) we had a lot of synchronization issues. Unfortunately, they were never really resolved, but sleeps and acks were added until it worked... The underlying problem is, that simple serial has no real bidirectional handshaking. In contrast to AES SHA3 and KMAC operations take also longer. This might increase the effect.

PR #285, opened 2024-01-11
[capture, sha3, kmac] Seed software LFSR also in non-batch mode

vogelpi:
At least for KMAC, the software LFSR is used to mask initial keys in non-batch mode.

This got factored out of #280.


nasahlpa (approved):
Thanks @vogelpi.
Yes, CI is not failing due to these changes, it is because on the CI machine, sometimes, the ACK command gets lost during transmission. I'll fix it when working on resolving #283.

vogelpi:
Ah, now I understand. So this is actually primarily a CI issue. Thanks @nasahlpa .

PR #286, opened 2024-01-11
[cw305] Various fixes and improvements for CW305

vogelpi:
This PR contains a couple of commits for fixing / improving the support for the CW305 such as:
- Updating the example capture configuration (previous one was completely out of sync)
- Fixing instructions for generating binaries and bitstreams
- Updating binaries to re-enable random-batch captures
- Switching to random-batch captures in CI, adding a fvsr-key-batch job in CI


ci/azure-pipelines.yml
nasahlpa, line 1: Do we want to have more test for CW305 in CI? Maybe also do TVLA on CW305 captured traces? For TVLA, I've created an issue (#287) to track which tests do we want to add.

But we can address this in a separate PR, for now having these tests in CI looks fine.

ci/cfg/ci_aes_sca_random_cw305.yaml
```diff
@@ -24,21 +24,22 @@ waverunner:
 capture:
   scope_select: husky
   # scope_select: waverunner
-  num_traces: 500
+  num_traces: 10000
   show_plot: True
   plot_traces: 100
-  trace_db: ot_trace_library
-  # trace_db: cw
+  #trace_db: ot_trace_library
```
nasahlpa, line 30:
```suggestion
  # trace_db: ot_trace_library
```
Do be consistent with the other config files here in this directory, I would suggest keeping the whitespace.

capture/configs/aes_sca_cw305.yaml
```diff
@@ -0,0 +1,54 @@
...
+  target_type: cw305
+  fpga_bitstream: "../objs/lowrisc_systems_chip_englishbreakfast_cw305_0.1.bit"
+  force_program_bitstream: True
+  fw_bin: "../objs/aes_serial_fpga_cw305.bin"
+  # target_clk_mult is a hardcoded value in the bitstream. Do not change.
+  target_clk_mult: 0.1
+  target_freq: 10000000
+  baudrate: 115200
+  output_len_bytes: 16
+  #protocol: "ujson"
```
nasahlpa, line 11:
```suggestion
```
Do we maybe want to remove the ujson entry for now? As we currently only support simpleserial for CW305 targets, we might not want to give other users the impression that this is working?

ci/cfg/ci_aes_sca_random_cw305.yaml
```diff
@@ -24,21 +24,22 @@ waverunner:
 capture:
   scope_select: husky
   # scope_select: waverunner
-  num_traces: 500
+  num_traces: 10000
```
nasahlpa, line 27: The reason for using a relative low number of traces was to keep the CI execution time low. Is capturing 10000 traces still fast enough and do we need that many traces for a quick visual inspection whether traces are OK or not?

nasahlpa (approved):
Thanks for the PR, @vogelpi!
I've added some small comments.

ci/cfg/ci_aes_sca_random_cw305.yaml
vogelpi: SG

capture/configs/aes_sca_cw305.yaml
vogelpi: Good idea!

ci/cfg/ci_aes_sca_random_cw305.yaml
vogelpi: Fair point. In CI, the batch capture rate is around 1500 traces/s so it's fairly quick. However, I've now reduced it 4000, as this seems already sufficient for a specific TVLA (results will not be very meaningful though of course!). I plan to add some CI step for the specific TVLA in a follow-up PR, see other issue. And this step should use the traces captured here.

ci/azure-pipelines.yml
vogelpi: Thanks for opening the issue. Let's discuss over there.

PR #288, opened 2024-01-12
[capture] Remove unused cfg files

nasahlpa:
As now almost all capture scripts were ported to the simple capture scripts, it is time to remove the old, unsupported config files in `capture/configs` and `ci/cfg`.

As far as I see, ECDSA256 is still missing, so I'd suggest leaving the config files here for now. Also OTBN is not yet in the CI (will be done when uJSON port for OTBN is completed).


vrozic (approved):

vogelpi (approved):
LGTM, thanks @nasahlpa.

PR #290, opened 2024-01-12
[analysis] Trace DB support for CECA

### LINKED ISSUE #272: Port CECA script to new Trace DB

nasahlpa:
This PR enables support for the new trace database for the CECA script.

Closes #272.


nasahlpa:
@vogelpi OK to remove the old `cw/correlation-enhanced_collision_attack.py` script? This script was already replaced by the `cw/ceca.py` script, which now was extended in this PR to support the new trace DB format and moved to `analysis/`.

vogelpi:
> @vogelpi OK to remove the old `cw/correlation-enhanced_collision_attack.py` script? This script was already replaced by the `cw/ceca.py` script, which now was extended in this PR to support the new trace DB format and moved to `analysis/`.

Hi @nasahlpa , yes it's okay to remove that one.

Thanks for the PR, I'll review it once we sure both versions result in comparable performance. Currently, I am running more trace captures with masking off for this purpose.

ci/ci_data/ceca/aes_cw310.db
nasahlpa, line 1: @vogelpi this DB file for integrating a CECA attack into CI is quite large (~60mb). Do you think it is OK to have such a large file in the repository?

Alternatively, we could capture traces during the CI job and then use the traces for the attack. However, we need quite some traces (>16k) for the attack to work and we want to have a consistent output in the CI. So I would prefer using pre-captured traces in the CI.

vogelpi: Hm, I think having the 60 MB trace file is maybe the better option as this would allow for reproducible CI. It definitely needs to go to Git LFS though.

The question is whether what do we want to keep testing in CI:
- The attack itself including # key byte deltas recovered successful key recovery?
  - With newer bitstreams we might need more fewer/traces or it might become completely impossible in the future. It really depends on the HW.
- That the attack can keep reading the data base format? Then we also do with fewer traces.

capture/project_library/project.py
```diff
@@ -101,16 +102,23 @@ def get_waves(self, start: Optional[int] = None, end: Optional[int] = None):
         """ Get waves from project.
         """
         if self.project_cfg.type == "cw":
-            return self.project.waves[start:end]
+            if (start is not None) and (end is not None):
+                return self.project.waves[start:end]
+            elif start is not None:
+                return self.project.waves[start]
+            else:
+                return self.project.waves
```
vogelpi, line 110: Question: why don't you need to handle the case where `end` is provided but not start? You now seem to then always provide the full data set.

capture/project_library/project.py
```diff
@@ -120,13 +128,28 @@ def get_plaintexts(self, start: Optional[int] = None, end: Optional[int] = None)
...
+        if self.project_cfg.type == "cw":
+            if (start is not None) and (end is not None):
+                return self.project.textouts[start:end]
+            elif start is not None:
+                return self.project.textouts[start]
+            else:
+                return self.project.textouts
+        elif self.project_cfg.type == "ot_trace_library":
+            return self.project.get_ciphertexts(start, end)
+
```
vogelpi, line 152: Is this new function related to fixing array slicing? Or is it maybe in the wrong commit?

capture/project_library/ot_trace_library/trace_library.py
```diff
@@ -47,11 +47,13 @@ class TraceLibrary:
     """
```
vogelpi, line 47: I think this change should be factored out into a separate PR.

It would be also good if you could provide some more details on why the database should not be held in memory for multi-processing support. At least to me, it's not obvious why this helps.

ci/azure-pipelines.yml
```diff
@@ -212,6 +212,30 @@ jobs:
...
+      apt update
+      apt install git-lfs
+    displayName: "Install system dependencies"
+  - bash: |
+      git-lfs pull
+    displayName: "Pull LFS binaries"
+  - bash: |
+      set -e
+      pushd ci
+      ../analysis/ceca.py -f ci_data/ceca/aes_cw310.db -d output -s 3 -w 4 -a 1060 1119 -n 16000
```
vogelpi, line 236: Have you tried this with the CW305 or CW310? There should be a huge difference between the two. BTW, all traces I provided to you were taken from the CW305.

analysis/ceca.py
```diff
@@ -17,6 +18,11 @@
 import ray
 import scared
 
+ABS_PATH = os.path.dirname(os.path.abspath(__file__))
```
vogelpi, line 21: I think it would be useful to have a comment here specifying why this is needed.

vogelpi (commented):
Thanks for working on this @nasahlpa , the changes to the CECA script seem incredibly simple, I think that the result of a well thought through project library :-)

I have mostly some question around the other commits. In my view, these should be better factored out into a separate PR.

capture/project_library/project.py
nasahlpa: If somebody just calls `get_waves(0)` (i.e. give me the wave at DB entry 0) then start is set to 0 as it is the first parameter. This is equivalent to `get_waves(start=0)`. End is in this case None.
I don't think that anybody will try `get_waves(end=0)`.

If this sounds like a non-optimal decision from my side, I could throw an error if somebody tries to do `get_waves(end=0)`

ci/azure-pipelines.yml
nasahlpa: Oh nicely spotted. Indeed, this trace dataset is from CW305 not CW310 - I'll fix that.

nasahlpa:
As suggested by @vogelpi, I've split up the PR into three different PRs.

- This PR now only adapts the ceca.py script such that it can handle traces captured with the new trace database.
- PR #293 fixes the opening and closing of DB files
- PR #294 fixes the array slicing bug

As this PR is based on the #293 and #294, we should first merge them before merging this one.

analysis/ceca.py
nasahlpa: Done.

capture/project_library/project.py
nasahlpa: You are right, it is related to this PR. So I've keep this here and moved the array slicing bug to a separate PR.

capture/project_library/project.py
vogelpi: Ah, now I understand. This sounds good to me. Thanks for explaining!

nasahlpa:
@vogelpi, this PR is now ready for review. It consists of two commits:
- Extend ceca.py script to also support the new trace DB
- Added CI job. 100 traces in the CW and OT trace DB are opened with the script. CI checks if the message "Failed to recover the AES key" is printed. This is the expected message as we only use 100 traces to save storage space here in the repo.

vogelpi (approved):
LGTM, thanks @nasahlpa !

PR #291, opened 2024-01-15
[capture] Set num_segments to max

### LINKED ISSUE #277: [capture] Auto-compute number of segments for Husky
In the past, we where automatically computing the number of segments per batch capture to maximize capture performance. Now, we always need to manually specify this value in the config file. But as different value are supported depending on the capture mode (random-batch vs. fvsr-key-batch) the user has to change multiple values in the config file.

I suggest to automatically choose the max value if no value is specified in the config file for Husky. WaveRunner we should probably leave untouched.

nasahlpa:
If no num_segments is provided in the config file for Husky, the scope driver automatically sets it to num_segments_max.

For WaveRunner, an error is raised if no num_segments max is provided.

Closes #277.

### DIFF
--- a/capture/capture_aes.py
+++ b/capture/capture_aes.py
@@ -112,15 +112,21 @@ def setup(cfg: dict, project: Path):
 
     logger.info(f"Initializing scope {scope_type} with a sampling rate of {cfg[scope_type]['sampling_rate']}...")  # noqa: E501
 
+    # Determine if we are in batch mode or not.
+    batch = False
+    if "batch" in cfg["test"]["which_test"]:
+        batch = True
+
     # Create scope config & setup scope.
     scope_cfg = ScopeConfig(
         scope_type = scope_type,
+        batch_mode = batch,
         acqu_channel = cfg[scope_type].get("channel"),
         ip = cfg[scope_type].get("waverunner_ip"),
         num_samples = cfg[scope_type]["num_samples"],
         offset_samples = cfg[scope_type]["offset_samples"],
         sampling_rate = cfg[scope_type].get("sampling_rate"),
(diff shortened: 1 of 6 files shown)


vogelpi (approved):
Thanks @nasahlpa for the PR. This looks great and importantly:
- WaveRunner isn't touched by this.
- We can still explicitly specify values for num_segments which are then left untouched.
- num_segments for non-batch mode is automatically chosen.

This improves the productivity when working with CW :-)

vogelpi:
One question: when doing batch-mode captures with CW I've noticed that the API reports to do 1 segment per batch which isn't true. Have you noticed this, too? Do you maybe know how to correct this?

nasahlpa:
Thanks for having a look into the PR!

Do you mean the following message:
`Connected to ChipWhisperer (num_samples: 501, num_samples_actual: 501, num_segments_actual: 1)`
If yes, we had this message already before working on this PR.
The reason for this message is that we first configure the ChipWhisperer and in a second step configure the batch mode together with the number of segments. However, the message you are seeing is generated in the first step. So I don't think that there is a way of changing this.

vogelpi:
> Thanks for having a look into the PR!
> 
> Do you mean the following message: `Connected to ChipWhisperer (num_samples: 501, num_samples_actual: 501, num_segments_actual: 1)` If yes, we had this message already before working on this PR. The reason for this message is that we first configure the ChipWhisperer and in a second step configure the batch mode together with the number of segments. However, the message you are seeing is generated in the first step. So I don't think that there is a way of changing this.

Yes, that's exactly what I meant. Sounds good to me then, thanks for your explanation. :-)

PR #293, opened 2024-01-15
[db] Fix DB opening and closing

nasahlpa:
This PR fixes issues that could occur when accessing the DB using multiple processes:
- When opening the DB, check if the DB already exists.
- When closing the DB, properly close the connection to the DB engine.

Moreover, also check if the user accidentally did not provide the DB file path with the .db extension. If this happens, simply add it to the path.

### DIFF
--- a/capture/project_library/ot_trace_library/trace_library.py
+++ b/capture/project_library/ot_trace_library/trace_library.py
@@ -47,11 +47,16 @@ class TraceLibrary:
     """
     def __init__(self, db_name, trace_threshold, wave_datatype = np.uint16,
                  overwrite = False):
+        # If .db extension is not provided, add it to the file.
+        if not db_name.endswith(".db"):
+            db_name = db_name + ".db"
+        # If overwrite flag is set, delete existing DB.
         if overwrite:
-            trace_lib_file = Path(db_name + ".db")
-            trace_lib_file.unlink(missing_ok=True)
-        self.engine = db.create_engine("sqlite:///" + db_name + ".db")
-        db_utils.create_database(self.engine.url)
+            Path(db_name).unlink(missing_ok=True)
+        # Create or open database.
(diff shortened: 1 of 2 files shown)


capture/project_library/ot_trace_library/trace_library.py
```diff
@@ -47,11 +47,16 @@ class TraceLibrary:
     """
     def __init__(self, db_name, trace_threshold, wave_datatype = np.uint16,
                  overwrite = False):
+        # If .db extension is not provided, add it to the file.
+        if ".db" not in db_name:
```
andreaskurth, line 51:
```suggestion
        if not db_name.endswith(".db"):
```
I think this would be more explicit and prevent obscure problems for which a `.db` suffix would get added unexpectedly (e.g., when a user provides `my.dbase` as `db_name`)

andreaskurth (commented):
One improvement suggestion, otherwise LGTM. Thanks @nasahlpa :+1:

capture/project_library/ot_trace_library/trace_library.py
nasahlpa: Ah yes, thanks for suggestion!

andreaskurth (approved):
LGTM, thanks @nasahlpa :+1:

capture/project_library/project.py
```diff
@@ -72,7 +72,7 @@ def close(self, save: bool) -> None:
         if self.project_cfg.type == "cw":
             self.project.close(save = save)
         elif self.project_cfg.type == "ot_trace_library":
-            self.project.flush_to_disk()
+            self.project.close(save = save)
```
vogelpi, line 75: Nit: now that the handling of the two databases is more aligned, the `if / else if` construct could be dropped.

vogelpi (commented):
LGTM! Thanks for factoring this out of the previous PR.

PR #294, opened 2024-01-15
[db] Fix bug in array slicing

nasahlpa:
Previously, retrieving parts of a dataset (e.g., plaintext, ciphertext, ...) did not work as there was a bug in the array slicing logic.

Basically, if start and end or end was not provided, "None" is the default value. However, the check in the if condition should be `if (start is not None)` and not `if (start)`.

This PR fixes this issue.

Moreover, now it is also possible to select a single entry of a dataset, e.g., `get_plaintexts(0)` returns the plaintext at DB ID 0.

### DIFF
--- a/analysis/tvla.py
+++ b/analysis/tvla.py
@@ -524,7 +524,7 @@ def run_tvla(ctx: typer.Context):
                     keys = np.empty((num_traces_orig, 16), dtype=np.uint8)
 
                 if general_test is False:
-                    keys[:] = project.get_keys()[trace_start:trace_end + 1]
+                    keys[:] = project.get_keys(trace_start, trace_end + 1)
                 else:
                     # Existing KMAC trace sets use a mix of bytes strings and ChipWhisperer byte
                     # arrays. For compatiblity, we need to convert everything to numpy arrays.
--- a/capture/project_library/ot_trace_library/trace_library.py
+++ b/capture/project_library/ot_trace_library/trace_library.py
@@ -118,10 +118,17 @@ class TraceLibrary:
             The traces from the database.
         """
         self.flush_to_disk()
-        if start and end:
+        if (start is not None) and (end is not None):
+            # SQL ID starts at 1.
+            start = start + 1
             query = db.select(self.traces_table).where(
                 (self.traces_table.c.trace_id >= start) &
                 (self.traces_table.c.trace_id <= end))
+        elif (start is not None):
+            # SQL ID starts at 1.
+            start = start + 1
+            query = db.select(self.traces_table).where(
+                self.traces_table.c.trace_id == start)
         else:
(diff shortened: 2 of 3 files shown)


capture/project_library/ot_trace_library/trace_library.py
```diff
@@ -118,10 +118,17 @@ def get_traces(self, start: Optional[int] = None,
             The traces from the database.
         """
         self.flush_to_disk()
-        if start and end:
+        if (start is not None) and (end is not None):
+            # SQL ID starts at 1.
+            start = start + 1
```
vogelpi, line 123: Don't you need to do the same for `end`? Otherwise one trace less than requested is returned, no?

vogelpi (approved):
LGTM, with two questions. Thanks for factoring this out @nasahlpa !

capture/project_library/ot_trace_library/trace_library.py
nasahlpa: The reason for this is the following:
In Python, `array[0:100]` gives you 100 elements from index 0 to 99.
With the new `get_plaintexts/ciphertexts/waves(0, 100)` functions, we want to imitate this behavior.
As the new trace DB index starts at 1 and not 0, we need an offset of 1, so we are adding +1 in the code.
However, as we still only want 100 elements in this example, we do not add this offset to the end. So we get elements 1 to 100.

analysis/tvla.py
```diff
@@ -524,7 +524,7 @@ def run_tvla(ctx: typer.Context):
                     keys = np.empty((num_traces_orig, 16), dtype=np.uint8)
 
                 if general_test is False:
-                    keys[:] = project.get_keys()[trace_start:trace_end + 1]
+                    keys[:] = project.get_keys(trace_start, trace_end + 1)
```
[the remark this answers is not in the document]
nasahlpa: No this should work for both database formats.
If you have a look into the `get_keys(start, end)` function in project.py, you'll find the following for the CW backend:
https://github.com/lowRISC/ot-sca/blob/643c5d3f4e7ea3f69e4bd5591f91ba300810af0e/capture/project_library/project.py#L118
So, in both cases, i.e., the CW or the new trace DB, we just get the key slice we requested

nasahlpa:
Thanks for your review, @vogelpi!

analysis/tvla.py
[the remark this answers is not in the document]
vogelpi: A I see, this makes sense!

capture/project_library/ot_trace_library/trace_library.py
vogelpi: Thanks for explaining, all good then :-)

PR #295, opened 2024-01-22
Aes fvsr data

vrozic:
Add support for capture aes-fvsr-data


nasahlpa:
Thanks for your work, I've copied your changes into #366 as I did not manage to rebase (missing permissions).

PR #297, opened 2024-01-23
[capture] Update uJSON binaries

nasahlpa:
As the KMAC, SHA3, and AES SCA applications are collected in the same uJSON binary, this PR removes the individual uJSON binaries used so far and replaces them with the combined sca_ujson_fpga_cw310 binary.  In addition, this update is needed as some binaries previously did not had access to the SW trigger, which is needed for tests on discrete. This binary was generated from OpenTitan commit: [b2d9b8356aa72164fb828e60ca8118495df6669d](https://github.com/lowRISC/opentitan/commit/b2d9b8356aa72164fb828e60ca8118495df6669d) with the command:
`bazel build //sw/device/tests/crypto/cryptotest/firmware:firmware`

### DIFF
--- a/ci/cfg/ci_kmac_sca_random_cw310_simpleserial.yaml
+++ b/ci/cfg/ci_kmac_sca_random_cw310_simpleserial.yaml
@@ -3,7 +3,7 @@ target:
   fpga_bitstream: ../objs/lowrisc_systems_chip_earlgrey_cw310_0.1_kmac_dom.bit
   force_program_bitstream: False
   fw_bin: ../objs/kmac_serial_fpga_cw310.bin
-  # fw_bin: "../objs/kmac_ujson_fpga_cw310.bin"
+  # fw_bin: "../objs/sca_ujson_fpga_cw310.bin"
   # target_clk_mult is a hardcoded value in the bitstream. Do not change.
   target_clk_mult: 0.24
   target_freq: 24000000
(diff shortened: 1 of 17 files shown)


ci/cfg/ci_kmac_sca_random_cw310_simpleserial.yaml
```diff
@@ -33,7 +33,7 @@ husky:
   scope_gain: 26
 capture:
   scope_select: husky
-  num_traces: 1000
+  num_traces: 100
```
vogelpi, line 36: Is this reduction in number of traces on purpose or accidental?

vogelpi (commented):
Thanks @nasahlpa for the PR, and especially for referencing the precise commit used for generating the binary. This allows us to exactly reproduce the binary in case we need to :+1:

ci/cfg/ci_kmac_sca_random_cw310_simpleserial.yaml
nasahlpa: Although it is not related to this PR, I noticed that capturing 1000 traces in the non-batch mode takes quite some while. As we do not need to capture that many traces, I suggest to reduce it to 100 traces.

vrozic: I agree it's a good idea to reduce it to 100. Some captures run at ~10 traces/s so capturing 1000 traces significantly slows down CI

vrozic (approved):
Thanks @nasahlpa for making the changes.
I suggest amending the getting started guide to include the correct bazel command for making the new binaries.

nasahlpa:
> Thanks @nasahlpa for making the changes. I suggest amending the getting started guide to include the correct bazel command for making the new binaries.

Thanks, I'll added this information to the getting started guide.

ci/cfg/ci_kmac_sca_random_cw310_simpleserial.yaml
nasahlpa: I've moved this into a separate PR #298

PR #298, opened 2024-01-23
[ci] Reduce number of traces

nasahlpa:
This PR reduces the number of traces to be recorded from 1000 to 100 for non-batch tests. For testing whether capture works, this is still enough and it speeds up CI.


vogelpi (approved):
LGTM, thanks for factoring this out!

PR #299, opened 2024-01-24
Prepare repo for multi-target support

nasahlpa:
Currently, the repository only supports a single SCA/FI target, i.e., CW FPGA (CW310/CW305). This PR adds initial support for enabling multiple targets (e.g., the discrete chip).

The following changes were made:
- Move `capture/lib/ot_communication.py` to `target/communication` and split into multiple files.
- Add a new driver class `target/targets.py` that supports multiple targets. Currently, only CWFPGA is support. In a subsequent PR, support for the discrete chip will be added.


target/targets.py
```diff
@@ -0,0 +1,123 @@
...
+    baudrate: Optional[int] = None
+    port: Optional[str] = None
+    read_timeout: Optional[int] = 1
+
+
+class Target:
+    """ Target class.
+
+    Represents a SCA/FI target. Currently, either a ChipWhisperer FPGA board
+    or the discrete chip are supported.
```
vogelpi, line 35: Ulta nit: this PR doesn't yet add support for the discrete chip ;-)

vogelpi (approved):
Thanks @nasahlpa , this LGTM!

Personally, I also think the splitting of the commands into individual files makes a lot of sense.

I am in favor of merging this quickly and before the other open PRs. Otherwise there is again a big risk that some of the other changes get lost during the re-organization this PR here performs.

target/targets.py
nasahlpa: Thanks, updated.

PR #300, opened 2024-01-25
[capture] Unblock Husky USB communication after power-up

vrozic:
CW Husky often blocks USB communication after power-up. 
This is causing problems with CI whenever Huskies are power-cycled.

Initializing the scope twice seems to solve the problem.


vogelpi (commented):
Thanks for the PR @vrozic . Are we sure this is getting the CI pipeline up again after powering off the Huskies?

nasahlpa (approved):
Thanks @vrozic, hope that this will fix the issue.

PR #302, opened 2024-01-26
Add support for the OpenTitan chip

nasahlpa:
This PR adds support for flashing, resetting, and communicating with the discrete OpenTitan chip.

### DIFF
--- a/target/chip.py
+++ b/target/chip.py
@@ -1,0 +1,53 @@
+# Copyright lowRISC contributors.
+# Licensed under the Apache License, Version 2.0, see LICENSE for details.
+# SPDX-License-Identifier: Apache-2.0
+"""OpenTitan chip utility functions."""
+import time
+from subprocess import PIPE, Popen
+from typing import Optional
+
+
+class Chip():
+    """Class for the discrete chip. Initializes OpenTitan with the provided
+    firmware & provides helper functions.
+    """
+    def __init__(self, firmware, opentitantool_path,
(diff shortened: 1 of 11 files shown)


target/chip.py
```diff
@@ -0,0 +1,50 @@
...
+                               "--exec", "transport init",
+                               "--exec", "bootstrap " + self.firmware, "no-op"],
+                              stdout=PIPE, stderr=PIPE)
+        flash_process.communicate()
+        rc = flash_process.returncode
+        if rc != 0:
+            raise RuntimeError('Error: Failed to flash OpenTitan chip.')
+        else:
+            # Wait until chip finished booting.
+            time.sleep(1)
```
vrozic, line 32: How sure are we that 1s is enough to finish booting?

nasahlpa: Good point - I've made some experiments and I think that 1s should be more than enough. For FI experiments, we might even reduce it to 500 or 750ms. I am changing the code such that the user can specify the boot delay.

vogelpi (approved):
This looks good, thanks @nasahlpa .

I think it would make sense to mention the exact commit you used to build the opentitantool binary. Writing something like https://github.com/lowRISC/OpenTitan/commit/ca93a12 should work (meaning you don't need to specify the branch name and full hash, 7 chars is fine) the important thing is that the commit is really upstreamed.

nasahlpa:
Thanks for your comments, I've included the commit hash and command used to build the opentitantool.

PR #269, opened 2024-01-04
[capture] FPGA mode for KMAC

nasahlpa:
This PR updates the uJSON interface to KMAC by transmitting the fpga_mode flag during initialization. This status flag is used by the target code to determine whether KMAC should run using FPGA specific features helpful for SCA or not. The changes for the OpenTitan SCA code are in #[20772](https://github.com/lowRISC/opentitan/pull/20772).

Issue #256, opened 2023-12-11
[uJSON] 


nasahlpa:
Currently, a small timeout ([`time.sleep(0.01)`](https://github.com/lowRISC/ot-sca/blob/e20ffa3f0210cf8b568d6bae89f59e1b6587bcf3/capture/lib/ot_communication.py#L43)) between uJSON commands are needed to prevent that the receive buffers overflow and the message gets lost.
To speed up capture rate, we should get rid of this problem.

PR #303, opened 2024-01-29
[capture] Add trigger source command

nasahlpa:
This PR adds the SCA trigger source command to the capture scripts. Needed to start penetration tests at ASIC as there is no HW trigger available.


target/communication/sca_trigger_commands.py
```diff
@@ -0,0 +1,40 @@
...
+        Args:
+            trigger_source:
+                            - 0: Precise, hardware-generated trigger - FPGA only.
+                            - 1: Fully software-controlled trigger.
+        """
+        if self.simple_serial:
+            self.target.write(cmd="t", data=bytearray([trigger_source]))
+        else:
+            # TODO: without the delay, the device uJSON command handler program
+            # does not recognize the commands. Tracked in issue #256.
```
vrozic, line 32: nit: We usually use this format for TODO. Makes it easier to search the repository.

`TODO(#256): Without the delay, the device uJSON command handler program does not recognize the commands.`

nasahlpa:
The uJSON tests are currently failing as this PR needs the changes from #269 and #304.

target/communication/sca_trigger_commands.py
vogelpi: Good point raised by @vrozic. Please make sure to align the format before merging.

target/communication/sca_trigger_commands.py
```diff
@@ -0,0 +1,40 @@
...
+            time.sleep(0.01)
+            self.target.write(json.dumps("TriggerSca").encode("ascii"))
+            # DisableMasking command.
```
vogelpi, line 35: This comment seems wrong.

vogelpi (commented):
Looks good module two comments, thanks @nasahlpa !

target/communication/sca_trigger_commands.py
nasahlpa: Fixed, thanks!

target/communication/sca_trigger_commands.py
nasahlpa: Hm I followed the recommendation of the OpenTitan [C style guide](https://opentitan.org/book/doc/contributing/style_guides/c_cpp_coding_style.html#todo-comments) regarding comments. There seems not to be a recommendation for Python though.

But I am following @vrozic advise and change it to `TODO(#256)` for now and I created an issue that we can streamline this later on.

PR #304, opened 2024-01-29
[capture] Add FPGA mode to SHA3

nasahlpa:
This PR enables transmitting the fpga_mode bit to the target. With this bit, the user can select whether SHA3 in the FPGA specific mode or the normal mode is started.


nasahlpa:
The uJSON tests are currently failing as this PR needs the changes from #303.

capture/capture_sha3.py
```diff
@@ -157,6 +157,14 @@ def configure_cipher(cfg, target, capture_cfg) -> OTSHA3:
     # Create communication interface to OT PRNG.
```
vogelpi, line 157: Didn't know where to put this comment but please note that this commit now references a PR of another repo using the short notation, i.e., `#No`. This is not working properly but may lead to issues in the future once there is an issue or PR in ot-sca with the number `No`. That's why I always recommend to use the more explicit notation `org/repo#pr` when adding references in commit messages.

For details see https://docs.github.com/en/get-started/writing-on-github/working-with-advanced-formatting/autolinked-references-and-urls .

vogelpi (approved):
LGTM!

Please try to use the explicit references in the future when possible.

PR #306, opened 2024-01-31
[db] Fix database reading

nasahlpa:
This PR fixes a bug when reading keys from the database. Thanks @vrozic for finding this!


vrozic (approved):
Thanks @nasahlpa for the quick fix

PR #307, opened 2024-02-01
[tvla] Enable re-plotting of figures

### LINKED ISSUE #284: [tvla] Loading `ttest_step_file` for replotting figures is broken
The `--tttest-step-file` option allowing to load a previously stored results file for regenerating plots is broken. It requires several variables and files (e.g. `single_trace.npy`, `num_traces`) which are not / no longer stored are no longer loaded.

This feature is important especially for the highly compute intense tests (specific AES testing) where generating the t-test step file takes several hours on a beefy workstation and printing the figures is a matter of seconds.

nasahlpa:
This PR adds metadata to the ttest_step_file that is needed to re-printing the figure.

A ttest_step_file can be created by setting passing the `--save-to-disk-ttest` argument to the tvla script. Make sure that `number_of_steps > 1`. To re-generate the figures, the ttest_step_file, which is stored in `tmp/ttest-step.npy.npz`, can be loaded using the argument `--ttest-step-file`.

Closes #284

### DIFF
--- a/analysis/tvla.py
+++ b/analysis/tvla.py
@@ -668,7 +668,12 @@ def run_tvla(ctx: typer.Context):
 
     elif cfg["ttest_step_file"] is not None:
         # Load previously generated t-test results.
-        ttest_step_file = np.load(cfg["ttest_step_file"])
+        ttest_step_file = np.load(cfg["ttest_step_file"], allow_pickle=True)
+        metadata = ttest_step_file['metadata'].item()
+        sample_start = ttest_step_file['sample_start']
+        single_trace = ttest_step_file['single_trace']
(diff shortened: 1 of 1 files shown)


vogelpi (approved):
LGTM, thanks for working on this @nasahlpa !

Question does this work with both database versions?

vogelpi:
@nasahlpa , do we want to merge this?

nasahlpa:
> LGTM, thanks for working on this @nasahlpa !
> 
> Question does this work with both database versions?

I've tested it with the reference CW database stored in `ci/ci_data/ceca/aes_cw305_cw_db.cwp`.

PR #310, opened 2024-02-07
[sca] Fix ACK for uJSON SHA3

### LINKED ISSUE #283: [ci] SHA3 captures seem to be broken/flaky in CI

nasahlpa:
This PR should fix the 'Acknowledge error: Device and host not in sync' error message that occasionally occured in the CI for SHA3 measurements when using uJSON.

With this change, the _ujson_sha3_sca_ack waits now until the ACK flag is received. If the device is, for some reason, really not responding, the readline() function times out and the exception is raised.

Closes #283


vrozic (approved):
LGTM

PR #311, opened 2024-02-07
[capture] Fix scalar vs array data type

nasahlpa:
This PR changes 1-byte array to scalar fields.
Binary was created using:
./bazelisk.sh build //sw/device/tests/crypto/cryptotest/firmware:firmware in lowRISC/opentitan@114749d


vrozic (approved):
LGTM

PR #315, opened 2024-02-09
[capture] Fix SHA3 FVSR

nasahlpa:
This PR fixes a bug in the SHA3 data generator that returned a non-fixed message for the SHA3 FVSR non-batch mode.

### DIFF
--- a/util/data_generator.py
+++ b/util/data_generator.py
@@ -107,7 +107,6 @@ class data_generator():
         ct = [x for x in ct_bytes]
 
         del (sha3_fixed)
-        self.advance_fixed()
         return pt, ct, key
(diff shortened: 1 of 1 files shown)


util/data_generator.py
```diff
@@ -107,7 +107,6 @@ def get_sha3_fixed(self):
         ct = [x for x in ct_bytes]
 
         del (sha3_fixed)
-        self.advance_fixed()
         return pt, ct, key
```
vrozic, line 110: Post-merge, sorry.

It looks like `key` output is never used in `capture_sha3.py`, so we could probably get rid of it here.

PR #316, opened 2024-02-09
[capture] Reset device on no response

nasahlpa:
When no response is received from the device, restart it and try again. This is helpful to avoid that a single missed response from the device requires us to throw away the entire capture results.

Currently only implemented for SHA3 to test the code.


vrozic (approved):
Thanks @nasahlpa for making this change.
The code looks good to me.

PR #319, opened 2024-02-14
[fi] Rename attack configs

nasahlpa:
As soon first FI scripts for the chip will be added, this PR, similar to the scripts in `capture/config`,  appends the cw310 suffix to the FI attack scripts located in `fault_injection/configs`. The reason for having `cw310 `and `chip` scripts is that the config files for these two targets are quite different (e.g., the target entry uses different fields). Having separate config files for each targets allows the user to more conveniently switch between the targets.


vogelpi (approved):
This looks good to me but I am wondering why this change is needed in the first place. Would you mind explaining please?

nasahlpa:
Sure, I've extended the PR description to explain the reasoning behind this PR.

vogelpi:
Thanks @nasahlpa , this makes sense. I think I must have been confused a bit: the PR mentions scripts but it actually changes the configs. I agree that these will be quite different for the different targets.

PR #320, opened 2024-02-15
[scope] Set WaveRunner short command header

nasahlpa:
When using the WaveStudio GUI to access the scope, this application reconfigures the format of the command responses to CHDR=OFF, i.e., there is no header. However, the Waverunner driver we are using requires that the commands are acknowledged in the header using CHDR=SHORT.


vogelpi (approved):
This looks sensible to me.

PR #321, opened 2024-02-15
[capture] Implement 12- and 16-bit LeCroy support

nasahlpa:
This PR enables to select the resolution of the attached LeCroy oscilloscope, i.e., 8-, 12-, or 16-bit.

### DIFF
--- a/capture/scopes/waverunner/waverunner.py
+++ b/capture/scopes/waverunner/waverunner.py
@@ -72,7 +72,7 @@ class WaveRunner:
         num_samples: Number of samples per segment.
     """
 
-    def __init__(self, ip_addr):
+    def __init__(self, ip_addr, bit):
         """Inits a WaveRunner.
 
         Connects to the oscilloscope, populates and prints device information.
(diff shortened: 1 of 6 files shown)


capture/scopes/waverunner/waverunner.py
```diff
@@ -397,8 +410,15 @@ def capture_and_transfer_waves(self):
...
         waves = self._parse_waveform(data)
-        # Put into uint8 range
-        waves = waves + 128
+        # Put into uint8/uint16 range.
+        if self.bit == 8:
+            waves = waves + 128
+        elif self.bit == 12:
+            # Scale from 16 to 12 bit. This is needed as the 12-bit oscilloscope
+            # returns a 16-bit value per sample.
+            waves = np.int16(waves / 65536 * 4096) + 2048
```
vogelpi, line 419: From a performance viewpoint is there any difference between the division and multiplication and simple shift right and left again? I would assume that numpy handles this in a smart way but it might be worth checking.

vogelpi (approved):
LGTM with one question. Thanks @nasahlpa !

capture/scopes/waverunner/waverunner.py
nasahlpa: I did the experiment and both versions of doing this seem to be similar in terms of performance.

PR #323, opened 2024-02-16
[capture] Re-add force_program_bitstream to target_cfg

vogelpi:
This parameter got somehow dropped and was no longer read from the capture config file. As a result, we were always using the same bitstreams. For example, the SHA3 measurements produce the following CI artifact:

bokeh_plot_faulty

Whereas the correct plot would look like this (generated locally using the "same" bitstream and binary):

bokeh_plot_correct

The active portion of the trace is 4x longer as the masked SHA3 takes 4 cycles per round vs. 1 cycle for the unmasked implementation.

### DIFF
--- a/capture/capture_aes.py
+++ b/capture/capture_aes.py
@@ -98,6 +98,7 @@ def setup(cfg: dict, project: Path):
         protocol = cfg["target"]["protocol"],
         pll_frequency = cfg["target"]["pll_frequency"],
         bitstream = cfg["target"].get("fpga_bitstream"),
+        force_program_bitstream = cfg["target"].get("force_program_bitstream"),
         baudrate = cfg["target"].get("baudrate"),
         port = cfg["target"].get("port"),
         output_len = cfg["target"].get("output_len_bytes")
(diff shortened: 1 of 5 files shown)


capture/capture_aes.py
```diff
@@ -98,6 +98,7 @@ def setup(cfg: dict, project: Path):
         protocol = cfg["target"]["protocol"],
         pll_frequency = cfg["target"]["pll_frequency"],
         bitstream = cfg["target"].get("fpga_bitstream"),
+        force_program_bitstream = cfg["target"]["force_program_bitstream"],
```
nasahlpa, line 101:
```suggestion
        force_program_bitstream = cfg["target"].get("force_program_bitstream"]),
```
I think we should use here the `.get()` functionality. The reason is that in the config files for the chip, we do not have the `force_program_bitstream` entry. Hence, when using `cfg["target"]["force_program_bitstream"]` the script would crash.

When using  `cfg["target"].get("force_program_bitstream"])` and the `force_program_bitstream` entry is not available (i.e., for the chip config), None is returned.

nasahlpa (approved):
Thanks for fixing this!

Could you please also apply the fix to the FI script:
https://github.com/lowRISC/ot-sca/blob/2c7f52dcd7880a0a073c1ee279a67894b7c590e7/fault_injection/fi_ibex.py#L45

capture/capture_aes.py
vogelpi: Good point, let me do that! And thanks for the explanation. I wasn't sure what the exact purpose of the `.get` was.

vogelpi:
Thanks @nasahlpa  for the review and feedback. I've now also adjusted `fi_ibex.py`.

PR #324, opened 2024-02-16
[tvla] Optimize memory usage

nasahlpa:
This PR optimizes the memory consumption of the TVLA script by freeing some memory objects. By setting an array to None, the Python internal garbage collector takes care of actually freeing the memory. With this PR, I've managed to reduce the mem. consumption of certain parts of the script to -50%.

### DIFF
--- a/analysis/tvla.py
+++ b/analysis/tvla.py
@@ -371,13 +371,12 @@ def run_tvla(ctx: typer.Context):
                                     )
         project = SCAProject(project_cfg)
         project.open_project()
-        proj_waves = project.get_waves()
         metadata = project.get_metadata()
 
         if cfg["input_histogram_file"] is None:
-            num_samples = len(proj_waves[0])
+            num_samples = len(project.get_waves(0))
         else:
-            assert num_samples == len(proj_waves[0])
+            assert num_samples == len(project.get_waves(0))
 
         if cfg["input_histogram_file"] is None:
             adc_bits = 12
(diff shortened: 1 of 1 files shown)


vrozic (approved):
Nice improvement.
The code looks good to me.

analysis/tvla.py
```diff
@@ -461,9 +461,8 @@ def run_tvla(ctx: typer.Context):
                 # Make sure to re-open the CW project file as we close it during
                 # the operation to free up some memory.
                 if i_step > 0:
-                    if not OTTraceLib:
-                        project.open_project()
-                        proj_waves = project.get_waves()
+                    project.open_project()
+                    proj_waves = project.get_waves()
```
vogelpi, line 465: For the OT trace database it might make sense to only get the waves you really need here, i.e., to call `get_waves()` with the specific arguments. But I might be wrong and the traces are only loaded when you actually access them. At least for the CW trace database this was the case as the waves are memory-mapped under the hood.

vogelpi (approved):
Thanks @nasahlpa this looks good. One question on potentially improving the performance / memory usage further.

analysis/tvla.py
nasahlpa: Good point, I've made the change. Could you please carefully check whether my changes are OK? I really want to avoid that we have a mismatch between traces and ptx/keys, e.g., trace[i] does not belongs to plaintext[i]/key[i]

nasahlpa: Tested with a couple of databases and I get the same results.

vogelpi (approved):
Thanks @nasahlpa , the code changes look good to me but I don't have the time for testing right now. Since you tested this thoroughly at your end, I am happy for you to proceed and merge this. Thanks!

PR #327, opened 2024-02-23
[fi] Add OTBN scripts

nasahlpa:
This PR adds the scripts for performing fault injection attacks on OTBN.

The binary was created from lowRISC/opentitan#21642


johannheyszl (approved):
Thanks for adding this @nasahlpa!

PR #328, opened 2024-02-26
[chip] Add support for multiple chips

nasahlpa:
This PR adds support for having multiple chips attached to the computer. By providing the serial number of the chip using the "usb_serial:: entry in the config files, ot-sca is capable of programming the correct device.


johannheyszl (approved):
thanks @nasahlpa lgtm

PR #329, opened 2024-02-29
[capture] Add FPGA mode for AES

nasahlpa:
On the FPGA, the AES core automatically raises the SCA trigger we are using for capturing traces. However, on the chip this functionality is disabled.

This PR transmits the fpga_mode bit to the code running on the FPGA or the chip. Based on this bit, the target code either manually raises the trigger or lets the AES core handle automatically the trigger.

The target code is available in https://github.com/lowRISC/opentitan/pull/21741 and the binary
sca_ujson_fpga_cw310 was created using this PR and the command:
./bazelisk.sh build //sw/device/tests/crypto/cryptotest/firmware:firmware

### DIFF
--- a/capture/capture_aes.py
+++ b/capture/capture_aes.py
@@ -188,6 +188,14 @@ def configure_cipher(cfg, capture_cfg, ot_aes, ot_prng):
         ot_aes: The communication interface to the AES SCA application.
         ot_prng: The communication interface to the PRNG SCA application.
     """
+    # Check if we want to run AES SCA for FPGA or discrete. On the FPGA, we
+    # can use functionality helping us to capture cleaner traces.
+    fpga_mode_bit = 0
+    if "cw" in cfg["target"]["target_type"]:
+        fpga_mode_bit = 1
+    # Initialize AES on the target.
+    ot_aes.init(fpga_mode_bit)
+
     # Configure PRNGs.
     # Seed the software LFSR used for initial key masking and additionally
     # turning off the masking when '0'.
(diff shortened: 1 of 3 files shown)


capture/capture_aes.py
```diff
@@ -187,6 +187,14 @@ def configure_cipher(cfg, capture_cfg, ot_aes, ot_prng):
         ot_aes: The communication interface to the AES SCA application.
         ot_prng: The communication interface to the PRNG SCA application.
     """
+    # Check if we want to run AES SCA for FPGA or discrete. On the FPGA, we
+    # can use functionality helping us to capture cleaner traces.
+    fpga_mode_bit = 0
+    if "cw" in cfg["target"]["target_type"]:
+        fpga_mode_bit = 1
+    # Initialize KMAC on the target.
```
johannheyszl, line 195: nit "KMAC"

johannheyszl (approved):
Thanks @nasahlpa, lgtm, one small q please: Any chance we can avoid this case distinction? Is it the same trigger GPIO pin? Or can we just leave the SW raising its trigger while the HW controls a different faster one?

capture/capture_aes.py
nasahlpa: Thanks, fixed.

nasahlpa:
> Thanks @nasahlpa, lgtm, one small q please: Any chance we can avoid this case distinction? Is it the same trigger GPIO pin? Or can we just leave the SW raising its trigger while the HW controls a different faster one?

Thanks for the review! As mentioned in [#21784](https://github.com/lowRISC/opentitan/pull/21784) this is currently not possible but we could do it in the future.

PR #332, opened 2024-03-04
[analysis] Trace aligning using mean trace

nasahlpa:
With this PR, the reference trace for the trace aligning function is determined by calculating the mean over num_traces_mean traces. The reason for that is that the trace aligning function achieves better results by using the mean as a reference trace.

### DIFF
--- a/analysis/trace_preprocessing.py
+++ b/analysis/trace_preprocessing.py
@@ -75,13 +75,14 @@ def parse_arguments(argv):
                         required=False,
                         default=100,
                         help="Number of traces to print")
-    parser.add_argument("-r",
-                        "--ref_trace",
-                        dest="ref_trace",
+    parser.add_argument("-n",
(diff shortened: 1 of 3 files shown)


vrozic:
If I understand the change correctly, option `--num_traces_mean` replaces the `--ref_trace`, so we will no longer have the possibility to specify the reference trace. For me this is OK, but the commit message should probably be edited to clarify this.

Also, could you change the `Trace preprocessing` section in the getting started guide to reflect this change?

nasahlpa:
Many thanks for the comment, @vrozic! Yes, you are right, the `--ref_trace` parameter is not available anymore. I've changed the commit message and adapted the documentation.

vrozic (approved):
Thanks @nasahlpa for making the changes.
The rest of the code looks good to me.

PR #334, opened 2024-03-06
[analysis] Fix num_traces in TVLA plot

nasahlpa:
Previously, when number_of_steps > 1, the TVLA plot only contained the number of traces used for a step, but not the total number of traces processed. This PR fixes the issue, now the total number of used traces is printed in the figure.

### DIFF
--- a/analysis/tvla.py
+++ b/analysis/tvla.py
@@ -457,6 +457,9 @@ def run_tvla(ctx: typer.Context):
             else:
                 sample_step_hist = 5
 
+        # Number of traces after filtering over all steps.
+        num_traces_used_total = 0
+
         for i_step in range(num_steps):
             num_traces = num_traces_vec[i_step]
             trace_start = trace_start_vec[i_step]
@@ -520,7 +523,6 @@ def run_tvla(ctx: typer.Context):
                 if ((save_to_disk_trace is True or save_to_disk_ttest is True) and
                         general_test and i_step == 0):
                     np.save('tmp/single_trace.npy', single_trace)
-
             else:
                 trace_file = np.load(cfg["trace_file"])
                 traces = trace_file['traces']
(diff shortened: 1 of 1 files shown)


analysis/tvla.py
```diff
@@ -457,6 +457,9 @@ def run_tvla(ctx: typer.Context):
             else:
                 sample_step_hist = 5
 
+        # Number of traces after filtering over all steps.
+        num_traces_used_total = 0
```
vrozic, line 461: I'm not sure that this value should always be zero.
The condition of this if branch is:
```
if (cfg["input_histogram_file"] is None or cfg["output_histogram_file"] is not None) \
            and cfg["ttest_step_file"] is None:
```
so it is possible that `cfg["input_histogram_file"] is not None` (e.g. when we specify both input and output histogram files),
so the initial value for `num_traces_used_total` should be the number of traces used in the input histogram file.

nasahlpa: Good point, thanks! However, do we actually read the number of traces of the histogram when `cfg["input_histogram_file"]` is provided? Should this be done here:
https://github.com/lowRISC/ot-sca/blob/6bbb5aca447b65a6a48f885a87f2f377a570ff88/analysis/tvla.py#L328

vogelpi: I think you're both somewhat right here. The histograms are defined as follows:
```
                # For every time sample we make two histograms, one for Hamming weight of the
                # sensitive variable = 0 (fixed set) and one for Hamming weight > 0 (random set).
                # histograms has dimensions [num_rnds, num_bytes, 2, num_samples, trace_resolution]
                # The value stored in histograms[v][w][x][y][z] shows how many traces have value z
                # at sample y, given that HW(state byte w in AES round v) = 0 (fixed set, x = 0) or
                # > 0 (random set, x = 1).
```
When an input histograms file is used, the number of traces used can be computed as sum(histograms[:][:][:][:][0][:]).

Since we already didn't set `num_traces_used_total` correctly when loading histograms before this PR. I suggest to create a new issue for tracking and merge this on as is.

vogelpi (approved):
LGTM, I suggest to create an issue for tracking the handling of the histogram loading feature.

analysis/tvla.py
vrozic: I agree that this can be merged as is.
I've added an item to issue #72 to track this.

PR #336, opened 2024-03-12
[fi] Optimize plot generation

nasahlpa:
This commit optimizes the generation of the FI scatter plot. Instead of calling plot.scatter() for each datapoint, now, an array is passed to the function. With this changes, scatter plots with >100k points can be plotted, which was not possible before this change.


johannheyszl (approved):
Thanks @nasahlpa

PR #337, opened 2024-03-12
[analysis] Add multi-processing for trace aligning

nasahlpa:
This commit adds the possibility of using multiple jobs for aligning the traces.


vogelpi (approved):
LGTM, thanks @nasahlpa !

PR #338, opened 2024-03-13
[fi] Add OTBN command handlers

nasahlpa:
This commit adds command handlers for the following tests:
- otbn_key_sideload
- otbn_load_integrity

The device code is located in lowRISC/opentitan#21917. The binary was created with this PR.


johannheyszl (approved):
thanks @nasahlpa

PR #340, opened 2024-03-19
[fi] Fix array splicing in DB

nasahlpa:
Similar to lowRISC/ot-sca#294, this commit fixes the array splicing for the FI database. Now, only the selected DB entries are returned instead all of them.


vogelpi (approved):
LGTM!

PR #341, opened 2024-03-19
Add GitHub Actions FPGA workflow

jwnrt:
This PR adds a GitHub Actions workflow which performs the same tasks as the Azure pipeline that runs FPGA jobs.

This work is part of a move from Azure to GitHub Actions generally in OpenTitan.

I'm planning to leave the `ci/azure-pipelines.yml` file here for a short while in case we have any issues with the GitHub Actions workflow so that we can switch back easier. Once we know it's working well we can remove that configuration.

### DIFF
--- a/.github/workflows/fpga.yml
+++ b/.github/workflows/fpga.yml
@@ -1,0 +1,313 @@
+name: fpga
+
+on:
+  push:
+    branches:
+      - master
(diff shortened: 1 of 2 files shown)


nasahlpa (approved):
Thanks for working on this, LGTM!

PR #342, opened 2024-03-19
[scope] Configure WaveRunner from config file

nasahlpa:
This commit allows the user to configure the channels, timebase, and the trigger for a WaveRunner scope using the configuration files.

When providing the config, the scope gets automatically configured allowing for reproducible capture runs.


vogelpi (approved):
Nice, this looks very useful for repeatable measurements. Thanks @nasahlpa !

PR #343, opened 2024-03-19
[ci] Add Python caching

jwnrt:


jwnrt:
This did not seem to provide any benefit. Possibly because many of our dependencies are from git and don't get loaded from the cache? I'm not sure. Closing.

vogelpi:
@jwnrt can you maybe explain the problem you're trying to solve with this please? Pascal or I might then be able to help. Is this to speed up CI? We're loading some Python dependencies as tarballs from GitHub because they're big. In case we need to speed this up, we could include this in the container image and periodically rebuild the container image.

jwnrt:
Oh apologies, I'm actually using this pull request to test the PR. I don't have permissions to manually run the pipeline, so I'm just force pushing to this PR to trigger it again.

I originally opened it to see if the dependency download time could be reduced, but it's only a minute long so not a priority to optimize right now.

PR #346, opened 2024-03-21
[fi] Command handlers for Ibex tests

nasahlpa:
This commit adds the command handlers and configs for the following Ibex FI penetration tests:
- ibex.char.flash_read
- ibex.char.flash_write
- ibex.char.sram_read
- ibex.char.sram_write
- ibex.char.unconditional_branch
- ibex.char.conditional_branch

The device code is located in lowRISC/opentitan#22135 and the binary was compiled from this PR using:
./bazelisk.sh build //sw/device/tests/crypto/cryptotest/firmware:firmware_fpga_cw310_test_rom

### DIFF
--- a/target/communication/fi_ibex_commands.py
+++ b/target/communication/fi_ibex_commands.py
@@ -54,6 +54,60 @@ class OTFIIbex:
         time.sleep(0.01)
         self.target.write(json.dumps("CharMemOpLoop").encode("ascii"))
 
+    def ibex_char_flash_read(self) -> None:
+        """ Starts the ibex.char.flash_read test.
+        """
+        # IbexFi command.
+        self._ujson_ibex_fi_cmd()
+        # CharFlashRead command.
+        time.sleep(0.01)
+        self.target.write(json.dumps("CharFlashRead").encode("ascii"))
+
+    def ibex_char_flash_write(self) -> None:
+        """ Starts the ibex.char.flash_write test.
+        """
+        # IbexFi command.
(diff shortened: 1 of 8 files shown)


target/communication/fi_ibex_commands.py
```diff
@@ -54,6 +54,60 @@ def ibex_char_mem_op_loop(self) -> None:
...
         self.target.write(json.dumps("CharMemOpLoop").encode("ascii"))
 
+    def ibex_char_flash_read(self) -> None:
+        """ Starts the ibex.char.flash_read test.
+        """
+        # IbexFi command.
+        self._ujson_ibex_fi_cmd()
+        # CharFlashRead command.
+        time.sleep(0.01)
+        self.target.write(json.dumps("CharFlashRead").encode("ascii"))
```
johannheyszl, line 64: nit: I guess something like "IbexCharFlashRead" would help to prevent potential confusion (down the line but not really worth the effort changing the encoding :)

johannheyszl (approved):
thanks @nasahlpa !

target/communication/fi_ibex_commands.py
nasahlpa: Thanks for your comment! I think this might be redundant as you have to send these two commands to the code to get the test started:
```
IbexFi
CharFlashRead
```

johannheyszl: thanks, yes

PR #348, opened 2024-03-25
[fi] Add register file cmd handler

nasahlpa:
This commit adds the command handlers for the following FI tests:
- ibex.char.register_file_read
- ibex.char.register_file

The device code is located in lowRISC/opentitan#21359


johannheyszl (approved):
Thanks @nasahlpa

PR #351, opened 2024-03-28
[fi] Add support for multiple Huskies

nasahlpa:
This commit enables us to run multiple FI setups using different Huskies. For this, the serial number can be specified with the fisetup.fi_gear_sn parameter in the config.


nasahlpa:
Thanks for the review! Yes, if the `sn` field is not provided in the yaml file, the scripts still work.

PR #352, opened 2024-03-30
[util] Add DB converter tool

nasahlpa:
This commit adds a converter tool capable of converting SCAProject libraries (CW or ot_trace_library) into a zarr database. This can be useful when experimenting with other SCA libraries, e.g., [decryptofy/scarr](https://github.com/decryptofy/scarr).


vogelpi (approved):
Nice!

PR #354, opened 2024-04-03
[fi] Add ChipShouter and ChipShover support

nasahlpa:
This commit adds support for doing EMFI with the ChipShouter and the ChipShover XZY table.


vogelpi (approved):
LGTM, this is inline with the other FI gear. Thanks for your work on this @nasahlpa !

PR #356, opened 2024-04-04
[capture] Add HMAC SCA

nasahlpa:
This commit adds the possiblility of capturing traces when a HMAC operation is performed. Note that on the device, HMAC from the cryptolib is used.

The device code is located in lowRISC/opentitan#22427


johannheyszl (approved):
Thanks @nasahlpa ! (I have only browsed quickly)

PR #359, opened 2024-04-15
[fi] Add SRAM Write Static Test

nasahlpa:
This commit adds the handler for the ibex.char.sram_write_static test.


johannheyszl (approved):
Thanks @nasahlpa

PR #360, opened 2024-04-15
Read device ID from target

nasahlpa:
This commit reads the device identifier from the target, which is transmitted after issuing a init command.


johannheyszl (approved):
Thanks @nasahlpa this will help the labs.

PR #362, opened 2024-04-18
Re-establish com interface after reset

nasahlpa:
This commit fixes the re-init of the COM interface after a reset was issued by the attack script.

### DIFF
--- a/target/targets.py
+++ b/target/targets.py
@@ -87,7 +87,7 @@ class Target:
         """Resets the target. """
         self.target.reset_target()
         if com_reset and self.target_cfg.protocol == "ujson":
-            self._init_communication()
+            self.com_interface = self._init_communication()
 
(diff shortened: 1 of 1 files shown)


target/targets.py
```diff
@@ -87,7 +87,7 @@ def reset_target(self, com_reset: Optional[bool] = False):
         """Resets the target. """
         self.target.reset_target()
         if com_reset and self.target_cfg.protocol == "ujson":
-            self._init_communication()
+            self.com_interface = self._init_communication()
```
johannheyszl, line 90: thanks. 
I was a bit puzzled but see that it was assigned in init only previousls. makes sense!

johannheyszl (approved):
Thanks @nasahlpa ! I know the labs will apreciate this improvemet :)

PR #364, opened 2024-04-24
[tvla] Optimize memory utilization

nasahlpa:
Previously, in i_step=0, the TVLA script opened all plaintext or keys at once. This is problematic, as for large datasets and large keys/ptx the memory consumption is tremendous.

This commit optimizes memory utilization by just opening the keys/ptx that correspond to the current step.


vogelpi (approved):
LGTM

PR #367, opened 2024-05-16
[fi/otp] Add OTP data read test to FI framework

wettermo:
This PR adds code for FI pen-testing the OTP module.

The test sequence can be summarized in the following way:
  1. Read content from a selected OTP partition for comparison values
  2. Perform fault injection on nop instructions
  3. Read content from the same OTP partition again
  4. Compare output with previous output and check if values differ

The OTP partition to read can be selected through the "which_test" parameter in the config file.
The corresponding PR in lowRISC/opentitan is https://github.com/lowRISC/opentitan/pull/23156


nasahlpa (approved):
LGTM, thanks!

PR #372, opened 2024-06-06
[fi] Update OTBN Key Sideload handler

nasahlpa:
This PR updates the OTBN Key Sideload hander to refect changes in the repo.


johannheyszl (approved):
thanks!

PR #373, opened 2024-06-14
[sca] Add KeySideloadFvsr and EcdsaP256 OTBN tests

nasahlpa:
This commit adds the following OTBN SCA routines to the code base:
- init: Init the OTBN SCA test on the device
- init_keymgr: Prepare the keymanager for the OTBN key sideloading test
- key_sideload_fvsr: Start the test
- read_response: Read response from device
- ecdsa_p256_sign: Test available in single, batch, and FvsR mode

Please note that this commit only adds these tests to the communication
stack but not to the capture script.

### DIFF
--- a/capture/capture_otbn.py
+++ b/capture/capture_otbn.py
@@ -21,7 +21,7 @@ from scopes.scope import (Scope, ScopeConfig, convert_num_cycles,
 from tqdm import tqdm
 
 import util.helpers as helpers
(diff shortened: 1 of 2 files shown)


nasahlpa:
Could I please get a review for this PR?

vogelpi (approved):
This looks good to me, thanks @nasahlpa !

PR #374, opened 2024-06-17
[fi] Add hardened check test

nasahlpa:
This commit adds the command handler for the
fi.ibex.char.hardened_check_eq_unimps test.


vogelpi (approved):
LGTM

PR #376, opened 2024-09-17
Add A1 bitstream and pentest firmware

andrea-caforio:
This commit adds the OpenTitan Earl Grey A1 FPGA bitstream and the pentest binary to the repository. The previous binaries for the Earl Grey Z1 version are retained with the filename suffix z1.

The FPGA bitstream was generated from commit https://github.com/lowRISC/opentitan/commit/282d863 and created with the command:
  `/bazelisk.sh build //hw/bitstream/vivado:fpga_cw310_test_rom`

The pentest binary was generated from commit  https://github.com/lowRISC/opentitan/commit/b2cee40 and created with the command:
  `./bazelisk.sh build //sw/device/tests/penetrationtests/firmware:firmware_fpga_cw310_test_rom`


vogelpi (approved):
Thanks @andrea-caforio and @nasahlpa for this PR! This looks good to me.

Two comments:
- It would be worth mentioning the in the commit message that the simple-serial based binaries are untouched and remain based on Z1. This is perfectly fine as we plan to remove these eventually.
- It's great that you mention the commit hashes used to build the the binaries and bitstreams. Please note however that in the ot-sca repo, opentitan hashes have no / a different meaning which can cause confusion. Please disambiguate using e.g. lowRISC/OpenTitan@282d863 . The GitHub UI automatically generates a link to the corresponding commit in the OpenTitan repo which helps a lot.

vogelpi:
Ah and of course, we need to debug the CI failures. I suspect there is an issue with the FPGA/ASIC mode for SHA3/KMAC. We've disables this feature in Earlgrey for the production tapeout such that the FPGA is again closer to the ASIC. See https://github.com/lowRISC/opentitan/pull/23157

nasahlpa:
Yes you are right, the tests are failing because of these parameters. In PR #378 I have updated the instruction how to build the bitstreams for the FPGA. I am currently building it for the CW310 and we test again :-)

nasahlpa:
Another question regarding the bitstreams @vogelpi, do you know why we have `lowrisc_systems_chip_earlgrey_cw310_0.1.bit `and `lowrisc_systems_chip_earlgrey_cw310_0.1_kmac_dom.bit` in the `objs/` folder? I couldn't find build instructions for the latter one.

vogelpi:
> Another question regarding the bitstreams @vogelpi, do you know why we have `lowrisc_systems_chip_earlgrey_cw310_0.1.bit` and `lowrisc_systems_chip_earlgrey_cw310_0.1_kmac_dom.bit` in the `objs/` folder? I couldn't find build instructions for the latter one.

The latter one has the masking for KMAC enabled. Doing this is not straightforward as you need to change the top-level to enough resources available. If you check the commit / PR where I've updated this bitstream, you should find a tag / branch of mine containing all the required commits.

For a more update to date version that also changes the clock ratios to have again 20x oversampling with Husky, refer to this branch: https://github.com/vogelpi/opentitan/commits/ot-sca-cw310-trivium/

nasahlpa:
Ah I see, thanks for the clarification. I've only checked the last git commit of the KMAC FPGA bitstream and there I couldn't find the tag / branch used for that - but an older commit had the information.

Not sure how we want to proceed with that. Relying on a OpenTitan fork for building this bitstream seems to be quite hacky, difficult for others to reproduce, and not so easy to keep the bitstream up to date with the master branch but on the other hand generating a new top just for the KMAC SCA measurements seems to be a bit too radical.

vogelpi:
I agree with your assessment but I also have an idea. Let's discuss offline on THU or FRI.

nasahlpa:
Closing this we've updated the binaries in #392.

PR #383, opened 2024-12-05
[fpga] Enforce CW310 SAM3X Firmware

nasahlpa:
As CW310 SAM3X firmware version below "1.5" have an UART flow control bug, enforce that the version is at leaset "1.5". This ensures that the communication over uJSON works flawless.


nasahlpa:
Unfortunately there seems to be a bug in the CW SAM3X FW. The binary was created on 24.04.24 and should have version number 1.5. However, when reading out the information from the FW on the board, I get:
```
import chipwhisperer as cw
target = cw.target(None, cw.targets.CW310)
print(target.latest_fw)
print(target.sam_build_date)
```

```
{'major': 1, 'minor': 2}
23:46:21 Apr 24 2024
```
The correct firmware from 24.04.24 is flashed but the version number still is the old one. Reported to NewAE - hence keeping this PR as a draft  until this is resolved.

PR #387, opened 2025-01-09
[chip] Add interface option

nasahlpa:
The voyager board uses the "ftdi" interface while other boards use the "hyper310" interface. This commit makes the interface parameterisable. By setting
```
target:
  interface: "ftdi"
```
in the corresponding FI/SCA config, ot-sca can be used to analyze
the voyager board.


vogelpi (approved):
LGTM.

Lint and pytest currently fail CI but it the errors don't seem related to this PR. Is it possible that we always use latest versions of dependencies and we're now seeing some incompatibility issue?

nasahlpa:
Thanks for the review, @vogelpi!

Yeah I assume something like this, tracked in #388.

Issue #388, opened 2025-01-14
Fix CI


nasahlpa:
Currently failing, investigate and fix it.

PR #389, opened 2025-01-16
[ci] Fix CI runners

nasahlpa:
Recently, the pytest and lintpy jobs were failing. The reason for this is that we are running these jobs on the latest Ubuntu version. However, we currently recommend using Ubuntu 22.04. Hence, lets pick this runner for the CI jobs.

Closes lowRISC/ot-sca#388


vogelpi (approved):
LGTM, thanks for investigating and fixing this Pascal!

PR #390, opened 2025-02-11
[cw] Add ID to CW Husky and CW310

nasahlpa:
Previously, attaching only one CW310 and Husky when not using a container was possible. With this commit, multiple boards and scopes can be connected when providing the ID of the Husky/CW FPGA board.

### DIFF
--- a/capture/capture_aes.py
+++ b/capture/capture_aes.py
@@ -103,7 +103,8 @@ def setup(cfg: dict, project: Path):
         port = cfg["target"].get("port"),
         output_len = cfg["target"].get("output_len_bytes"),
         usb_serial = cfg["target"].get("usb_serial"),
-        interface = cfg["target"].get("interface")
+        interface = cfg["target"].get("interface"),
+        husky_serial = cfg["husky"].get("usb_serial")
     )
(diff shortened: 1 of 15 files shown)


nasahlpa:
FYI @h-filali

vogelpi (approved):
LGTM but I have one question: if we don't specify a serial number (because a user has just one board connected) does this still work?

nasahlpa:
Thanks for the review!

Yes, if you do not provide the serial number of the husky or the CW board (like in the CI run), no serial number is passed to the underlying newaetech/chipwhisperer driver. If only one board is connected, the driver will take this board. If multiple boards are connected, an error message is printed by the driver.

This works because I am using the `.get()` to receive the serial number:
https://github.com/lowRISC/ot-sca/blob/8107fc127fd5446d30e6c37a2a8dcbd022f689a8/capture/capture_aes.py#L105

if none is provided (there is no such entry in the cfg file), `None` is returned by `.get()`. This `None`  then is forwarded to the CW driver, which ignores it.

vogelpi:
Perfect, thanks for your explanation Pascal.

